PR #51, opened 2021-10-01
Update the tree view

marekweb:
The debug tree view displays the entire structure of the handbook. 

- Each page is shown in its position in the directory structure
- Each page's structure is shown based on its table of contents

Check out the page on the deploy preview here: (renamed to /meta):
- https://deploy-preview-51--sourcegraph-handbook.netlify.app/meta


src/styles/page.scss
```diff
@@ -135,3 +135,7 @@
     color: inherit;
     opacity: 0.8;
 }
+
+.italic {
```
felixfbecker: We already have `fst-italic` (**f**ont-**st**yle: **italic**) from Bootstrap :)

felixfbecker (approved):

src/styles/page.scss
marekweb: Nice, thanks!

PR #86, opened 2021-10-04
Document how to check links locally

jasonyavorsky:


README.md
```diff
@@ -22,6 +22,22 @@ Then start the handbook:
...
+find content -name '*.md' |xargs -n1 yarn run markdown-link-check -c .github/workflows/link-check-internal.json
+```
+
+This can be slow, so you can also check a single file by running this command, replacing `path_to_file` with the file you want to validate.
+
+```
+yarn exec markdown-link-check -c .github/workflows/link-check-internal.json <path_to_file>
+```
+
+The [configuration file](./.github/workflows/link-check-internal.json tells the link checker to not validate external links. If you wish to have them checked, you can remove the `-c .github/workflows/link-check-internal.json` portion of the command invocations above.
```
felixfbecker: We actually have a script in package.json for this, so the instructions can be simpler :)

I'd also remove the clutter from explicitly not checking external links when running on a single file. We really only need this in CI to make sure builds are reproducible. Locally it seems valuable to also check external links, at least when only looking at the file you're currently editing (not the entire handbook).

Also small nit: links should have a descriptive title for accessibility and avoid only saying "here".

```suggestion
We use [markdown-link-check](https://github.com/tcort/markdown-link-check) for link checking at build time in [the `link-check` GitHub action](.github/workflows/link-check.yml). If you want to run it locally, from the root of the repository you can run this command:

    yarn check-internal-links

This can be slow, so you can also check a single file by running this command, replacing `path_to_file` with the file you want to validate:

    yarn markdown-link-check <path_to_file>
    
Note that this will also check external links, which the GitHub action ignores. If you wish to ignore those, add `-c .github/workflows/link-check-internal.json` to the command.
```

felixfbecker:
Thanks for adding these ☺️

jasonyavorsky:
Thanks @felixfbecker !

PR #97, opened 2021-10-04
Update .nvmrc

umpox:
According to https://nodejs.org/en/download/releases/, Node v16.10.0 is no longer available. Changing this to the current latest version

Errors seen in this PR: https://github.com/sourcegraph/handbook/pull/96 and in the first commit on this branch


jasonyavorsky:
Grabbed a local copy, updated, and browsed around and things looked fine.

felixfbecker (approved):
Thanks!

Issue #7, opened 2021-09-30
Implement new left navigation sidebar


felixfbecker:
See design: https://www.figma.com/file/ze8C5qjzSTadKv2LYmON1T/Handbook-Wizardry?node-id=1144%3A3038

PR #101, opened 2021-10-04
Add left navigation

unclejustin:
This is really just a POC for a dynamic left navigation. It is in no way ready for use, but may be a good jumping off point for collaboration.

Feedback and contributions appreciated!

Quinn's design: https://www.figma.com/file/ze8C5qjzSTadKv2LYmON1T/Handbook-Wizardry?node-id=1144%3A1036

Related to: https://github.com/sourcegraph/handbook/issues/7

### DIFF
--- a/src/pages/[...slug].tsx
+++ b/src/pages/[...slug].tsx
@@ -1,13 +1,18 @@
+import { readdirSync, readFileSync, existsSync } from 'fs'
+import { join } from 'path'
+
 import { Toc } from '@stefanprobst/rehype-extract-toc'
 import { GetStaticPaths, GetStaticProps } from 'next'
 import ErrorPage from 'next/error'
 import Head from 'next/head'
+import Link from 'next/link'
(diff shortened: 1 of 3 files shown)


src/pages/[...slug].tsx
```diff
@@ -147,6 +181,41 @@ export const getStaticProps: GetStaticProps<PageProps> = async ({ params }) => {
...
+    const leftNav: LeftNav[] = topLevelPaths
+        .map(path => {
+            const pathToIndex = join(process.cwd(), CONTENT_FOLDER, path, 'index.md')
+
+            if (!existsSync(pathToIndex)) {
+                return
+            }
+
+            const contents = readFileSync(pathToIndex, 'utf8')
+            const heading = contents.match(/^# (.*)$/m)?.[1]
```
Comment (felixfbecker) [inline src/pages/[...slug].tsx:197]: I think we should integrate with the markdown rendering code instead, which already generates a ToC and would be more reliable

PR #106, opened 2021-10-04
Delete duplicate edit page

marybelzer:


felixfbecker:
Consider renaming the file to `index.md`, that way the URL will be `/editing` instead of `/editing/editing` 🙂

PR #113, opened 2021-10-04
fix individual direction pages link

benvenker:
Just had an extra `#` in the link. Removed.


jasonyavorsky:
@benvenker thanks so much! Feel free to just merge this sort of obvious fixes, I will never mind.

PR #125, opened 2021-10-05
update amplitude onboarding

ebrodymoore:


ebrodymoore:
fyi @attfarhan replacing how to add events with the denylist

PR #131, opened 2021-10-05
Rename theme to focus area

malomarrec:
Rename theme to focus area to match [rollup page](https://handbook.sourcegraph.com/direction/code-graph). It's better if we use the same wording everywhere.
If we want to use themes instead, then we should evolve the rollup page to match (the opposite of this)


jasonyavorsky:
@malomarrec if we want them to be the same, I'll change the roll-up.

PR #161, opened 2021-10-06
Properly attribute quote in Code graph principles

christinaforney:


jasonyavorsky:
Thanks @christinaforney for the fix and sorry for the problem! :heart:

PR #170, opened 2021-10-07
Engaging other teams update

virginiaulrich:
@Joelkw in follow-up to [our Slack chat](https://sourcegraph.slack.com/archives/C01HW836940/p1632170258225800), I added what we talked about in step 3 of the engaging other teams file and as a note in step 8 in the workflow file. 

There are other changes in the engaging other teams doc that is for another purpose, so feel free to ignore those.


content/support/engaging-other-teams.md
```diff
@@ -30,7 +48,10 @@ When you need help, post in our #customer-support-internal Slack channel first t
 
 ## Step 3: Identify which team can help you
 
-In many cases, questions can span multiple teams. For example, a question about how to scale up indexed search to serve a large set of repositories could cover the Distribution, Cloud, and Search teams. In such cases, start where you think is best and our teammates in engineering will help us figure out if we need to go to another engineering team.
+A couple of notes before to help you get started in figuring out which team will be best to help:
+
+- In some cases, questions can span multiple teams. For example, a question about how to scale up indexed search to serve a large set of repositories could cover the Delivery, Repo Management, and Search teams. In such cases, start where you think is best and our teammates in engineering will help us figure out if we need to go to another engineering team.
+- For features tagged beta or experimental in our docs, it’s okay to err on the side of engaging engineering sooner rather than later (since we don't yet know if we will want to release these, we keep our docs light in case we opt to remove the feature after all).
```
Joelkw:
```suggestion
- For features tagged beta or experimental in our docs, it’s okay to err on the side of engaging product + engineering sooner rather than later (since we don't yet know if we will want to release these, we keep our docs light in case we opt to remove the feature after all. Also, for these features we're still figuring out how users expect them to work, so often "bugs" are informative to understand product or education gaps in the feature – you're helping the team by surfacing them directly).
`
```
Two changes suggested, in order: 
1. I'm not sure if this is redundant, if you/your team generally uses "engineering" throughout these docs to necessarily include product managers.
2. This is true! I actually set slack up so I'm being notified about as many bugs/threads in #support- channels mentioning insights as I can, but I do sometimes miss them anyway. Even though I don't always engage, it's really helpful to know the details of the early bugs (often "not directly an eng bug, but customer expectations ≠ product works") in an effort to get ahead of them by GA release. But I'm okay if you feel it's best not to include this second sentence.

Joelkw:
Thank you for adding this @virginiaulrich !

Joelkw (approved):

virginiaulrich:
@Joelkw accepting your changes! Though, we do just say "engineering" to mean the team that also includes a PM ... we do that since 90% of the time, we work with an engineer when we engage for help. That said, fair to add product to that line given the context, so let's do it! But also want to explain why I won't make that change everywhere.

tamarj0y (commented):
Read through your proposed changes. Everything looks good.

PR #172, opened 2021-10-07
Add additional tips for working async

jasonyavorsky:


content/company/asynchronous-communication.md
```diff
@@ -121,6 +121,9 @@ Here are some tips and things you should keep in mind when working at Sourcegrap
...
+- **Share what changed**. If a decision was made async, it's important to get the word out. Although it can feel like oversharing, especially for a remote company it's important to repeat yourself and hit the important communication channels with the news. Otherwise, it's easy to miss.
+- **Rely on the handbook as the source of truth**. A lot of decisions we make (perhaps nearly all of them) should end up with updating the handbook. If all of us maintain the handbook, it will remain a wonderful resource to find out the latest thinking on any topic. If we let it go, people will lose trust, and it will no longer be an effective way to document decisions.
```
jasonyavorsky:
```suggestion
- **Rely on the handbook as the source of truth**. A lot of decisions we make (perhaps nearly all of them) should end up with updating the handbook. If all of us maintain the handbook, it will remain a wonderful resource to find out the latest thinking on any topic. If we let it go, people will lose trust, and it will no longer be an effective way to document decisions. If you see something out of date or that we no longer practice in the handbook, prioritize updating it.
```

jasonyavorsky:
cc @InesRoitman

PR #192, opened 2021-10-07
Add suggested max review waiting period for marketing and product director review

Joelkw:
cc @jyavorska @abschu – thanks for these updates! I would propose adding some time guidelines if we want to be clear about what needs a review and when. 

Feel free to massage/change the wording. I'm mostly focused on maintaining a "merge it if you haven't gotten a review after 3 days anyway" culture for what I perceive as "smaller" updates in these categories (main example that's come up for the extensibility and code insights teams before – we add videos/demos/documentation to the youtube channel) and don't want to add too many obstacles to moving quickly. I of course understand (and expect PMs know) that major site updates, press engagements, or huge packaging/pricing changes of course go through the review even if more time as needed. 

Perhaps an alternative way would be to split these items up into items that "required a completed review (and specifies how long marketing + product directors must be given to do it)" and items that "require a 'review opportunity' that gets forfeited if incomplete after three days have passed."


jasonyavorsky:
@Joelkw (cc @abschu) I'd say go ahead and merge, if there isn't any feedback. It seems like a relatively non-controversial change, nobody's goal is blocking.

PR #203, opened 2021-10-08
added almost all tables from the excel file. Excluded ones noted in the file. Still needs proofreading and review of concepts

adeola-ak:


amberfurbush0317 (commented):
Look good to me

PR #204, opened 2021-10-08
Minimize use of ambigious Code Graph term

jasonyavorsky:
Code Graph can refer to an org within Sourcegraph, or an area of the product. It was being used in a mixed way in the direction page and this PR fixes that.

Since this does not change any of the actual meaning of the content on the page, going to just merge without review.

### DIFF
--- a/content/direction/code-graph/index.md
+++ b/content/direction/code-graph/index.md
@@ -4,7 +4,7 @@ We have teams working on [Search](./search/index.md), [Code Insights](./code-ins
 
 ## Mission
 
-We build our product to make finding, understanding, and editing code easier, regardless of how much you have, how complex it is, where you store it, or even how technical you are.
+We want to make finding, understanding, and editing code easier, regardless of how much you have, how complex it is, where you store it, or even how technical you are.
 
 ## Vision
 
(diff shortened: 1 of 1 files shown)


content/direction/code-graph/index.md
```diff
@@ -1,10 +1,10 @@
...
+We have teams working on [Search](./search/index.md), [Code Insights](./code-insights/index.md), [Code Intelligence](./code-intelligence/index.md), and [Batch Changes](./batch-changes/index.md); this page describes the overarching mission and strategy that ties them together, but each linked team also provides a more detailed direction page with what they are working on next and why. There is also a whole-company [Sourcegraph direction](../index.md) available.
 
 ## Mission
 
-We build the Code Graph to make finding, understanding, and editing code easier, regardless of how much you have, how complex it is, where you store it, or even how technical you are.
+We build our product to make finding, understanding, and editing code easier, regardless of how much you have, how complex it is, where you store it, or even how technical you are.
```
jasonyavorsky:
```suggestion
We want to make finding, understanding, and editing code easier, regardless of how much you have, how complex it is, where you store it, or even how technical you are.
```

PR #255, opened 2021-10-12
exclude beta and experimental features from SLAs

tammy-zhu:
@virginiaulrich @Joelkw - this is an optional add for clarity. I added the exclusion in the beta and experimental features page: https://docs.sourcegraph.com/admin/beta_and_experimental_features. If you like it, keep it (or edit it). If not, fine to remove my addition.


virginiaulrich (approved):
I think it's good to call it out in both places

PR #260, opened 2021-10-12
Clean up anchors on strategy page

jasonyavorsky:
There was an old table of contents (TOCs are now auto-generated and appear on the right) as well as some non-working anchors; this cleans them up but does not otherwise change the content on the page.


jasonyavorsky:
Cc @malomarrec

PR #287, opened 2021-10-13
Combine direction page(s) into strategy page

jasonyavorsky:
This is a prospective PR to see if we like the idea of combining the direction and strategy pages into one, as (briefly) discussed at https://sourcegraph.slack.com/archives/C01SD4A49E1/p1634127673277000. 

I didn't change any content here, just structure. The combined page looks extra long because of the older existing one-year direction content which we will update/shrink after setting Q4 OKRs.


content/company/values.md
```diff
@@ -1,6 +1,6 @@
 # Sourcegraph values
 
-These values are the beliefs and principles that help us achieve our [goals](goals/index.md) and [vision](strategy.md#vision) and build an [inclusive team](../communication/code_of_conduct.md). These apply at every level of our organization—"you" refers to individuals, teams, and Sourcegraph as a whole.
+These values are the beliefs and principles that help us achieve our [goals](goals/index.md) and [vision](../direction/index.md#vision) and build an [inclusive team](../communication/code_of_conduct.md). These apply at every level of our organization—"you" refers to individuals, teams, and Sourcegraph as a whole.
```
jasonyavorsky:
```suggestion
These values are the beliefs and principles that help us achieve our [goals](goals/index.md) and [direction](../direction/index.md) and build an [inclusive team](../communication/code_of_conduct.md). These apply at every level of our organization—"you" refers to individuals, teams, and Sourcegraph as a whole.
```

content/direction/index.md
```diff
@@ -1,12 +1,74 @@
 # Sourcegraph direction
 
-Product strategy is the bridge (how we're going to get there) between our [vision](../company/strategy.md) (where we're headed) and the [individual team and org directions](index.md#per-area-direction-pages) (what we are going to do).
+Our overall purpose is we want to make it so **everyone can code**. A world where everyone, not just ~0.1% of the world population, can code will see faster and more broadly beneficial technological progress. This page describes how we are going to achieve that goal, and the [individual team and org directions](index.md#per-area-direction-pages) describe what we're doing in each of our product areas.
```
jasonyavorsky:
```suggestion
## Purpose

Our overall purpose is we want to make it so **everyone can code**. A world where everyone, not just ~0.1% of the world population, can code will see faster and more broadly beneficial technological progress. This page describes how we are going to achieve that goal, and the [individual team and org directions](index.md#per-area-direction-pages) describe what we're doing in each of our product areas.
```

sqs (approved):
I like the idea of merging the current strategy and direction pages because they're essentially the same kind of thing. However, I would prefer for them to be called "strategy" instead of "direction" after we merge them because:

- If they're called "direction", then when people ask about our strategy and I point them to our "direction" page, then I'll have to explain most times why we call it direction and not strategy
- "Strategy" to me (as a word) includes the problem and how we're solving it; direction doesn't necessarily include the problem. Of course, our direction pages do include the problem, but "strategy" is a better term for what these are.

jasonyavorsky:
> I like the idea of merging the current strategy and direction pages because they're essentially the same kind of thing. However, I would prefer for them to be called "strategy" instead of "direction" after we merge them because:
> 
> * If they're called "direction", then when people ask about our strategy and I point them to our "direction" page, then I'll have to explain most times why we call it direction and not strategy
> * "Strategy" to me (as a word) includes the problem and how we're solving it; direction doesn't necessarily include the problem. Of course, our direction pages do include the problem, but "strategy" is a better term for what these are.

Ah yeah, I could see that for sure. Strategy is a clearer, more generally used word than direction. I'll wait to implement that until all the feedback is in if we decided to go ahead, but that's certainly doable.

nicksnyder (approved):
I agree with unifying/merging this. Don't have a preference on what to call it

jasonyavorsky:
@sqs I integrated your suggestion and think it looks much better now! Thanks for that.

To everyone, I'm going ahead with merging this as it seems to be a clear improvement, if a big change. As always, feedback and further iterations are always welcome, and if I've managed to break anything please let me know and I'll jump up to fix it.

PR #291, opened 2021-10-13
Make link check a little quieter

jasonyavorsky:


felixfbecker (approved):
Looks great and I think we should merge it! Unfortunately I think it wouldn't change the verbosity of the GitHub Action. 

We _could_ change the GitHub Action probably very easily by just changing to run `yarn check-internal-links` instead of using the packaged GitHub Action.

(Or we can go for the solution I outlined in Slack but that'll take more time)

jasonyavorsky:
The GitHub action is a bit less important since you can search on the page, instead of trying to keep up with the lines scrolling by in your terminal. I'll just merge this for now, and we can delete it once the better solution is implemented.

felixfbecker:
Agree with merging! But I disagree with this statement:

> The GitHub action is a bit less important since you can search on the page, instead of trying to keep up with the lines scrolling by in your terminal.

90% of the people contributing to the handbook won't run the handbook locally or even (know how to) use the terminal, considering many people in the company (probably the majority) don't have a technical background. Nor should they – the online UI should provide a good contribution experience 🙂

jasonyavorsky:
What I meant @felixfbecker was that the experience was good on the web because you can search on the page, not that web users are less important.

PR #295, opened 2021-10-13
Update locations.geojson

david-sandy:
Added my location.


david-sandy:
The deploy version doesn't look right, the map is at the bottom of the screen, yet, the json I added looks correct.

andreeleuterio (approved):

PR #312, opened 2021-10-14
Updating workflow to clarify feature request section

virginiaulrich:
@benjaminwgordon how do these changes look?


benjaminwgordon (commented):
LGTM, the wording is much clearer on CE vs CSE responsibility for feature requests

virginiaulrich:
Thanks for taking a look! I swapped the order too to lead with the more common scenario. Cheers!

PR #346, opened 2021-10-15
Add code insights to the support page

coury-clark:


content/support/engaging-other-teams.md
```diff
@@ -137,6 +137,12 @@ The cloud team is a newly formed team at Sourcegraph focused on rapidly scaling
 
 Any questions about Sourcegraph cloud should be routed to the [Cloud SaaS](../engineering/cloud/saas/index.md) via the [#cloud-saas slack channel](https://sourcegraph.slack.com/archives/C025BKWFPDY).
 
+### Code Insights
+
+**Keywords**: `insights`, `visualization`, `tracking`, `migrations`, `measuring`, `technical debt`
```
Joelkw:
```suggestion
**Keywords**: `insights`, `visualization`, `tracking`, `measuring`, `technical debt`, `search results over time`
```

@coury-clark I worry that "migrations" will collide with the dev use of Sourcegraph migrations (I know that "how we do migrations" was a problem that kept taking down k8s? If customers will hit similar problems not sure if "migrations" are the right inclusion.)

coury-clark: that's a good call out, it just seemed like these keywords were generally more product level than technical. I'll remove it to avoid confusion

coury-clark:
```suggestion
**Keywords**: `insights`, `visualization`, `tracking`, `measuring`, `technical debt`
```

PR #390, opened 2021-10-18
Adding BCP for customer redlines

virginiaulrich:
@srj427 @tammy-zhu @eliRsourcegraph here is a draft meant to easily satisfy what our customer has specified in the MSA we have been talking about. What do you think?


virginiaulrich:
Review happened in Slack [here](https://sourcegraph.slack.com/archives/C02J296T69K/p1634575497007400?thread_ts=1634573809.004300&cid=C02J296T69K)

PR #401, opened 2021-10-18
Include extension version when filing issues

abeatrix:
Update internal process docs to make sure we include extension versions when filing issues with the extensibility team.
More context in https://sourcegraph.slack.com/archives/C01JR51JR5J/p1634592070217200?thread_ts=1634554821.142100&cid=C01JR51JR5J


virginiaulrich (approved):
LGTM; thank you for taking the time to clarify and update this!

PR #436, opened 2021-10-20
changing links to relative paths

InesRoitman:


felixfbecker:
Did it not work with the `./` prefix? 🤔

InesRoitman:
Sorry @felixfbecker! I saw your ./ after I did the changes (they appeared as conflicts and I resolved them by removing them haha)

felixfbecker:
Oh. I guess we both fixed paths to be relative at the same time. The `./` is optional, `.` just means "the current directory", so it can do with either `./` or without it.

PR #499, opened 2021-10-25
CS OKR updates

virginiaulrich:
This is still WIP; creating PR to give visibility while I still work on this


virginiaulrich:
I seem to be in a loop with prettier and I don't have the energy for it. I will try again in a new PR

virginiaulrich:
New PR https://github.com/sourcegraph/handbook/pull/534

PR #557, opened 2021-10-27
Removed redundant content

serinadawn:
There were a few places in the page that content was redundant.  Most specifically in the definition of an active user.  I collapsed this content and removed some redundant references throughout the page.


content/bizops/user_definitions.md
```diff
@@ -19,22 +19,19 @@ Users can be in a number of different states. We track the following categories
...
+An activated user is a user who has established the core habit of the product. An active user is differentiated from a visitor by performing a qualifying event, which are intended to represent someone who has actively **engaged with** and **received value** from the product; something that takes a user from their first Search to establishing a habit around the core value proposition of the app. Qualifying events are [listed as 'Active' in Amplitude](https://analytics.amplitude.com/sourcegraph/govern/project/333976/events?filter=live)
 
 Qualifying events are not intended to be difficult, or prove that someone is a heavy user of the product. They are carefully selected to represent activities that are important to realizing the value of Sourcegraph.
 
 Performing any single activity in the above list makes a user "active" for the time period being measured.
 
 ### What is a retained user?
 
-A user is retained if they were active within a measured period if they were active in both the current and previous periods.
+A user is retained if they qalified as an active user (as defined above) within a measured period if they were active in both the current and previous periods.
```
jasonyavorsky:
```suggestion
A user is retained if they qualified as an active user (as defined above) within both the current and previous periods.
```

content/bizops/user_definitions.md
```diff
@@ -44,8 +41,6 @@ Time periods we track:
 - Weekly active user (WAU)
 - Monthly active user (MAU)
 
-Monthly is the most common reference frame we use, although we also use weekly within quarterly OKRs to help ensure we are able to impact the metric within the time period.
```
jasonyavorsky: Did we already say why we use weekly vs monthly somewhere else?

jasonyavorsky (approved):

content/bizops/user_definitions.md
serinadawn: I took this out since it is technically no longer true.  I don't think at the moment we need to qualify what we use until we settle on Q4 KRs.  Based on our conversation the other day it seemed to me that WAU or the ratios of DAU/MAU are in fact the most important measures for our business.

jasonyavorsky: Is there anything we can say about why you might use one or the other that is true? It feels odd to list them and leave the rest for the reader to figure out on their own.

PR #567, opened 2021-10-27
Add Kelvin-Lee to list of CSEs with managed-instance access

Kelvin-Lee:


virginiaulrich (approved):
Thanks for making these changes!

PR #584, opened 2021-10-28
add AARRR experiment

rrhyne:
Adds a page to planning experimentations


benvenker (approved):
Looks great!

jasonyavorsky (approved):
Looks good, but rather than adding a page information on what we might do, might it be better to link to a PR to the planning page (https://handbook.sourcegraph.com/product/product_management/planning) that implements these changes?

rrhyne:
> Looks good, but rather than adding a page information on what we might do, might it be better to link to a PR to the planning page (https://handbook.sourcegraph.com/product/product_management/planning) that implements these changes?


@jyavorska I would rather not have an open PR sitting around for a month or two so I added the experimental section to that page. WDYT?

jasonyavorsky:
Oh that's fine @rrhyne I didn't realize there was a couple months left here.

PR #599, opened 2021-10-29
Correct feature data chart

Joelkw:
@muratsu @tjkandala  I think this needed these corrections, yeah? And do we support Phabricator in the extension or should that be removed?


Joelkw:
@muratsu @tjkandala can one of you confirm what's correct and then hit the big green button? :)

tjkandala (approved):
Looks right (so does Phabricator, anything with a native integration should work on the browser extension)

PR #625, opened 2021-11-01
Revised Delivery support process

dan-mckean:
Bringing the process by which Delivery offers support back in line with the way other teams work, with a support rotation.


virginiaulrich (approved):
This captures it correctly. I will update https://handbook.sourcegraph.com/support/engaging-other-teams to reflect there is no longer an exception in place for Delivery team

StephanX (approved):
LGTM

PR #630, opened 2021-11-01
Add getting started pattern

rrhyne:


rrhyne:
@sourcegraph/design LMK if you feel there is a better place for this type of design system documentation!

content/product/design/wildcard_design_system/index.md
```diff
@@ -65,3 +65,18 @@ When a change to an existing component is required, the designer who is proposin
...
+
+## User interface patterns
+
+This section will detail user interface patterns not defined in the Wildcard Design System.
+
+### Getting started pages
+
+Getting started pages are presented to viewers of major Sourcegraph feature pages when they have not seen the page or when there is no content for the section. For un-authenticated users, the page serves as an overview and prompts the user login or sign up. For authenticated users, the page serves the same purpose and also provides helpful links to start using the feature.
+
+The following rules should be applied when selecting the page to display to users:
```
limitedmage, line 77: Can you add a rule about the URL schema? Something like
> The standard feature home page and the getting started view should have the same URL. For example, both the batch changes list and the batch changes getting started page are available at `/batch-changes`. Clicking the tabs at the top switches between the standard home page and the getting started page without modifying the URL.

Without this rule, it's complicated to build this functionality because it's hard to know how the user got to the page and if they should be redirected to getting started or not.

rrhyne: Yes, thanks for the suggestion!

content/product/design/wildcard_design_system/index.md
```diff
@@ -65,3 +65,18 @@ When a change to an existing component is required, the designer who is proposin
...
+
+- If the user has not seen the feature, the feature should display the getting started page
+- If the user has seen the feature and there is content in the section, the page should display the index of feature content
+- If the user has seen the feature and there is no content in the section, the getting started page will display
+- If the feature is not available in the instance, it will display for all users at all times
```
rrhyne, line 83:
```suggestion
- If the feature is not available in the instance, it will display for all users at all times
- The feature home page and the getting started view should have the same URL. For example, both the batch changes list and the batch changes getting started page are available at /batch-changes. Clicking the tabs at the top switches between the standard home page and the getting started page without modifying the URL.
```

rrhyne: @limitedmage see the above suggestion. Should we specify that adding the anchor #getting-to the URL started could set the proper tab?

limitedmage: How would you know if navigating to the root of the feature should add `#getting-started` or not?

ETA:
It's pretty obvious when you navigate to the page externally (eg. from the top nav), you redirect to `#getting-started` if there are not items.
However, if you click on the home tab, what should happen? It's just a URL, so it redirects back to `#getting-started` if we follow the same logic.
We either have to keep track of where the link was clicked from, and do special handling when clicking on the tab, or add a `#home` to the URL when clicking on the tab, or not have the URL change at all between tabs.

limitedmage: Looks good, modified to add highlighting for the URL.
```suggestion
- If the feature is not available in the instance, it will display for all users at all times
- The feature home page and the getting started view should have the same URL. For example, both the batch changes list and the batch changes getting started page are available at `/batch-changes`. Clicking the tabs at the top switches between the standard home page and the getting started page without modifying the URL.
```

limitedmage (approved):

LawnGnome (approved):
This makes sense to me. :+1:

content/product/design/wildcard_design_system/index.md
```diff
@@ -65,3 +65,19 @@ When a change to an existing component is required, the designer who is proposin
...
+
+This section will detail user interface patterns not defined in the Wildcard Design System.
+
+### Getting started pages
+
+Getting started pages are presented to viewers of major Sourcegraph feature pages when they have not seen the page or when there is no content for the section. For un-authenticated users, the page serves as an overview and prompts the user login or sign up. For authenticated users, the page serves the same purpose and also provides helpful links to start using the feature.
+
+The following rules should be applied when selecting the page to display to users:
+
+- If the user has not seen the feature, the feature should display the getting started page
```
AlicjaSuska, line 79:
```suggestion
- If the user has not seen the feature, the feature should display the getting started page
- If the user has not seen the feature but there is content available in this section, a modified getting started page should be displayed
```

This will be the case for code insights that users would see the feature for the first time but there will already be content to display. We're still deciding on how to present it.

AlicjaSuska (approved):

content/product/design/wildcard_design_system/index.md
```diff
@@ -65,3 +65,19 @@ When a change to an existing component is required, the designer who is proposin
   - Modify the main instance of the component in Wildcard
   - Publish changes with a fitting commit message
   - Add to the "Changelog" layer in the Wildcard Figma file
+
+## User interface patterns
+
+This section will detail user interface patterns not defined in the Wildcard Design System.
```
jjinnii, line 71: @rrhyne 
Thoughts on having this information live in WDS in Figma, maybe by the getting started page design comp? I agree that this is the best place to put it given our existing team pages, but I personally would not have thought this information lived here (I would have checked Figma or even checked Figma project files for getting started).

content/product/design/wildcard_design_system/index.md
rrhyne: Awesome. what would the modifications to the getting started page be?

Also, I'm curious what content will exist in new installations?

content/product/design/wildcard_design_system/index.md
rrhyne: I think that's because we do not have design documentation anywhere but Figma.

@AlicjaSuska can you comment on where the WDS documentation project is currently?

PR #639, opened 2021-11-01
Update GraphQL Voyager instructions

gabtorre:


abeatrix (approved):
This is good, thank you so much

PR #652, opened 2021-11-02
Adding details about Hubspot form

virginiaulrich:


airamare01 (approved):
Makes sense and understood from the get go. I'd say we're all set for Hubspot

PR #661, opened 2021-11-02
Add tutorial session info

jasonhawkharris:


daxmc99 (approved):
Sounds good, might also be a good idea to link to a folder where recordings are stored in case everyone on this list is out-of-office.

PR #700, opened 2021-11-04
Changing titles

airamare01:
Following the change of title for CS folks, making the necessary changes to reflect this in this doc.


airamare01:
Deleting this since V is working on a PR to make changes to the titles on CS handbook

PR #725, opened 2021-11-04
Slight update to our workflow

topebali:


virginiaulrich (approved):
Huzzah! Great changes and thanks for fixing that typo too!

PR #777, opened 2021-11-08
Business metrics are internal to Sourcegraph

nicksnyder:
We should document the level of visibility of business metrics. I assume these should be internal and not public (hence the PR), but if this is not right then let's update this PR and document whatever is correct.


content/company/values.md
```diff
@@ -83,11 +83,11 @@ Being open and transparent builds trust with our customers and within our team.
...
+| Public                  | Anyone outside of Sourcegraph can access this information. **This is the default level of transparency.**                              | [Handbook](../index.md), [code](https://github.com/sourcegraph/sourcegraph), [issues](https://github.com/sourcegraph/sourcegraph/issues), public Google Docs such as [RFCs](../communication/rfcs/index.md) | Goals, plans, all code, product docs, dev docs, company docs                                               |
+| Internal to Sourcegraph | Information that is not public for specific reasons, but is shared with Sourcegraph teammates.                                         | Public Slack channels, Google Docs accessible to everyone at Sourcegraph, internal [Google Groups](https://groups.google.com/all-groups)                                                                    | Business metrics, board slides, company financial info, customer information, open security investigations |
```
jasonyavorsky, line 89: This is the actual current metric values from reporting, and also goals with relative or absolute numbers as under this heading?: https://handbook.sourcegraph.com/company/strategy/code-graph#make-the-power-of-our-features-easier-to-find-and-use

(i.e., in addition to not saying what our current WAU is, do not show that we are trying to increase WAU by 15%, or aiming for 3 customers to use a feature, etc.?)

nicksnyder: Good question. Personally, I am more comfortable talking about relative increases in things like WAU, but would not be comfortable talking about absolute numbers.

sqs: I agree that relative increases are more likely to be OK than absolute numbers.

@dadlerj Can your team come up with some guidance around this?

sqs (approved):

nicksnyder:
Going to merge this as-is, and agree that more specific guidance would be helpful to add.

content/company/values.md
jasonyavorsky: Relative vs. absolute would be really nice, clear guidance if that's workable. Something like "You should never publicly state the starting value in a metric. The target is fine as long as it is a %, and fixed increases (such as 'increase by 5') are at your discretion."

PR #788, opened 2021-11-09
remove old 1-year vision section, link to new FY23 strategy

sqs:
The 1-year vision is from last year and is getting outdated. This adds a link to the new FY23 strategy, which is not quite complete but is much more up to date than the other content. The FY23 strategy right now doesn't include the product vision but it will when it enumerates the specific use cases for us to target (which we are working with the team to select now).


christinaforney (approved):
🎉

PR #811, opened 2021-11-10
Search: groom top-level strategy page

lguychard:
- Link to individual team pages
- Remove redundant "Strategy"
- Make "guiding principles" section more compact, and remove mention of work items for each guiding principles: these risk becoming out-of-date and would be better listed on individual team strategy pages. Also emojis (not strongly attached to these, just felt like it :D)
- Remove "with the launch of Sourcegraph Cloud" paragraph from "where we are now", because it targets what's next rather than where we are, and it's not stuff we're directly addressing on Core or Product at the moment.
- Remove redundant "vision" section from team page in `/engineering`


benvenker (approved):
LGTM! Big fan of the emojis 💯

PR #821, opened 2021-11-10
Update CE laptop section 

amenne:
Updating CE laptop guidance to 32GB of RAM requirement.


amenne:
@maureenlsourcegraph correcting CE guidance based on email thread. Since Apple doesn't make 14.2 with 32GB removed its reference and marked 16.2 as the standard, similar to CS.

maureenlsourcegraph (approved):
Approved from Tech Ops!

PR #828, opened 2021-11-10
Add talent review schedule

nicksnyder:
We need to commit to a schedule for talent review moving forward.

Given that we want to do it quarterly, there are roughly three options:
- First month of the quarter
- Second month of the quarter
- Third month of the quarter

I am proposing that we do it in the first month of each quarter for the following reasons:
- Second month of the quarter would land in December in Q4, and that seems like a bad month from a scheduling perspective.
- Third month of the quarter is when we are going to be planning for the next quarter, so don't want to overlap important processes like that.
- First month by process of elimination and I can't think of any downsides.

Since we just did a cycle, the next cycle would be Feb 2022.


chrispine (approved):
LGTM, thanks!

yinquanteo (approved):
Lgtm

jeanduplessis (approved):

PR #829, opened 2021-11-10
Clarify temporary projects

daxmc99:
Adding every temporary project as a GCP project is an excessive burden. We should only add **permanent** projects to the infra folder.


kevinwojo (approved):
Thanks for adding the clarification! 

LGTM

PR #833, opened 2021-11-11
Update dependent care for travel

DebbieSG:
Updated as per Desene's request to remove the reference to budget (there isn't one) and indicate that it needs to be discussed upfront.


cassiemelani:
Adding @desenesterling for her review and @mercadon would also love your thoughts since we've talked at length about this policy! Thank you both.

desenesterling (approved):

mercadon (approved):

PR #845, opened 2021-11-11
Fix file path for support

abeatrix:
Fix file path for support page


DaedalusG (approved):
LGTM! Thanks for adding an index and fixing some of those bad links

Kelvin-Lee (approved):
For all files with diffs, checked the edited links.

Edited links seem to go where they are intended to.

PR #881, opened 2021-11-15
Demo Repo Instructions

shawnplusplus:
Instructions for editing and maintaining documents in the demo repo.


marybelzer:
@shawnplusplus I'm working on implementing a check to the Handbook that requires all pages are linked to from somewhere. I noticed this page is not linked to from anywhere in the Handbook, and that the page is blank. Is this page still needed? If so, where would be the best place to link from? took my best guess [here](https://github.com/sourcegraph/handbook/pull/998) but let me know if i should move it.

PR #910, opened 2021-11-16
Updating managed instance config requests

emchap:
These are out of date! Updating to add reference to the Accounts repo instead of Hubspot and make it clear that Delivery doesn't need to set up the license key for the user; also updating to reflect that CE makes the Issue.


kevinwojo (approved):
Thanks! LGTM

PR #927, opened 2021-11-17
Use marketing site description for Batch Changes

jasonyavorsky:


jasonyavorsky:
Up to you if this is better or your prefer it as-is/something else @malomarrec

malomarrec:
@jyavorska I actually think the current description is better. The marketing page was a compromise between targeting enterprise buyers and developers. For this page, I think we care more about conciseness and clarity than addressing multiple audiences.
I'm going to close this for now, feel free to re-open if needed.

jasonyavorsky:
@malomarrec its up to you. Maybe there's a smaller adjustment? That it finds code you want to change and changes it by running code feels almost too vague. That could be describing almost anything that processes code.

PR #928, opened 2021-11-17
Fix indents

jasonyavorsky:
Only the top level item should be `#`, and then every other section should add one `#` per level of indent. This should make the table of contents behave correctly.


jasonyavorsky:
@malomarrec I'm going to merge this because I don't think it's changing any meaning on the page, but please feel free to change something back if anything was intentional.

PR #933, opened 2021-11-17
Make table of contents always show h1 h2 h3

### LINKED ISSUE #901: H3s always visible in generated page outline
Right now H3s only appear in the side outline when you scroll to that section. I would find it helpful if they always appeared.

jasonyavorsky:
Resolves https://github.com/sourcegraph/handbook/issues/901 (maybe)

This makes it so the full table of contents, including the h1 (`#`) page title, h2 (`##`) and h3 (`###`) headings, are always visible.

It makes the font size slightly smaller and never wraps the heading text so that it is more likely to fit on the page, but if there are lots and lots of headings it could result in scrolling on the right. That was possible before, but could be more likely now.

I think it still looks nice, and doesn't have the popping in and out which IMHO looks better. You can see it in action on any of the pages in the review app (https://deploy-preview-933--sourcegraph-handbook.netlify.app/)

In the future, if we want to get more clever, we could hide contributors in cases where the table of contents needs the space. But don't know at this point if that's even really necessary.

As a side effect this seems to fix the issue where pages "vibrate" on narrow screens.

### DIFF
--- a/src/pages/[...slug].tsx
+++ b/src/pages/[...slug].tsx
@@ -51,44 +51,6 @@ const HEADING_SELECTOR = 'h1, h2, h3, h4, h5, h6'
 export default function Page({ page }: PageProps): JSX.Element {
     const markdownBodyReference = useRef<HTMLElement>(null)
     const tocReference = useRef<HTMLElement>(null)
-    useEffect(() => {
-        const observer = new IntersectionObserver(
-            entries => {
-                for (const entry of entries) {
-                    const tocLink = tocReference.current?.querySelector<HTMLAnchorElement>(
-                        `a[href="#${entry.target.querySelector(HEADING_SELECTOR)!.id}"]`
-                    )
-                    const listItem = tocLink?.parentElement
-                    if (entry.intersectionRatio > 0) {
-                        // Visible: Mark <li> item and all ancestors (<li>s, <ul>s) as active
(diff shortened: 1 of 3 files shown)


jasonyavorsky:
Cc @quinnkeast @felixfbecker @marybelzer @nicksnyder what do you think?

felixfbecker:
I'd say let's have the discussion on the issue (#901)

src/pages/[...slug].tsx
```diff
@@ -46,49 +46,9 @@ export interface PageProps {
     page: PageWithMetadata
 }
 
-const HEADING_SELECTOR = 'h1, h2, h3, h4, h5, h6'
-
 export default function Page({ page }: PageProps): JSX.Element {
     const markdownBodyReference = useRef<HTMLElement>(null)
     const tocReference = useRef<HTMLElement>(null)
-    useEffect(() => {
```
Comment (felixfbecker) [inline src/pages/[...slug].tsx:54]: Imo even if we only show H2, we shouldn't remove the highlighting (bolding) of the current section. Imo it's a nice anchor of where you are in the document, like the "YOU ARE HERE" dot on a public map. Especially useful when you followed a link to a specific section.

jasonyavorsky:
Closing this since it sounds like someone else is going to work on it in a better way soon.

PR #944, opened 2021-11-18
Create communications page

msennett22:
Creating a communications page to outline Product/Engineering communication channels


content/product-engineering/communication.md
```diff
@@ -0,0 +1,28 @@
+## Quarterly Product/Eng All Hands Meetings
```
nicksnyder: This page needs to have a title (starts with `#` not `##`).

Either this can be a general communication page, or you could name the title/file specifically about all-hands.

PR #946, opened 2021-11-18
Move product and engineering under product-engineering

nicksnyder:


nicksnyder:
`Check that there are no large binary files` is failing but I am just moving files that already exist, so going to ignore this and merge anyway.

PR #820, opened 2021-11-10
Document product and engineering planning process

msennett22:
This creates a new section in the handbook, `product-engineering`, where we can document things that impact product and engineering equally (like planning), and then it adds documentation for our new planning process. The plan is to move the current `product` and `engineering` folders under this new folder, but that will not happen in this PR.

This PR includes and obsoletes the content in [this doc](https://docs.google.com/document/d/1vYMH3s7B8mP2jRx5_9qYOiVJLs2ZX6T1i9NFHM1xgGk/edit#).

PR #948, opened 2021-11-18
Bring in content from deleted planning page

jasonyavorsky:
This moves some content that was deleted from the old product-only planning page as part of https://github.com/sourcegraph/handbook/pull/820 to new locations. The prioritization page seemed like a reasonable home.


jasonyavorsky:
FYI @rrhyne this moves some of the content you made.

PR #952, opened 2021-11-18
CI playbook: replace gravity -> severity

pdubroy:


jhchabran (approved):
Thank you for proofreading! 

Was my previous wording implying that incident are pulled toward the ground with varying forces?

pdubroy:
> Was my previous wording implying that incident are pulled toward the ground with varying forces?

😆  No, it's perfectly reasonable in "normal" English (e.g., a common phrase is "the gravity of the situation"). It's just that _severity_ is the more commonly used term when referring to incidents.

PR #965, opened 2021-11-18
incidents: structure guide, update docs, add archive channel

bobheadxi:
- gives some structure to the incident guide, remove duplicate information, and simplify headings for ease of reference
- some outdated information here and things that should be in playbooks - moved things around a bit
- add step to archive channel

| before | after |
| ------ | ---- |
| image | image  |


jhchabran (approved):
Thank you for updating this 🚀

indradhanush (approved):

PR #977, opened 2021-11-19
Add more members to team info page

jasonyavorsky:
This adds more members to the teams (design, pmm, em). 

It also fixes a bug with relative paths caused by the introduction of the `product-engineering` top-level heirarchy.

### DIFF
--- a/data/features.yml
+++ b/data/features.yml
@@ -2,7 +2,7 @@
 
 literal_search:
   title: Literal Search
-  product_team: search
+  product_team: search_product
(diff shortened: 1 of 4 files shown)


jasonyavorsky:
@lguychard @benvenker important thing for you to be aware of here - there was just one search team listed, when in fact there are two. I split them out, but just assigned all the features to `search_product`. If you want something different please make a follow-up PR.

benvenker:
thanks!

PR #990, opened 2021-11-19
Clarify tracker update expectations

nicksnyder:
Previous documentation asked for an explicit weekly update for each issue in the tracker. There are a lot of issues in the tracker, some of which might not have a meaningful update each week.

This PR documents that we just expect issues have up-to-date information on a weekly basis and directors can set more specific expectations for their orgs as desired.

### DIFF
--- a/content/product-engineering/planning-process.md
+++ b/content/product-engineering/planning-process.md
@@ -68,27 +68,12 @@ There are a few ways we track plans and celebrate progress at Sourcegraph.
 
 ### GitHub project tracker
 
-The [GitHub project tracker](https://github.com/orgs/sourcegraph/projects/214) is **the source of truth** for key results in product and engineering. There are a few tabs.
+The [GitHub project tracker](https://github.com/orgs/sourcegraph/projects/214) is **the source of truth** for key results in product and engineering.
+
+There are a few tabs:
 
 - The **Department KRs** tab tracks the top level product and engineering key results that the VPs and Directors review every week at the [VP team sync](#vp-team-sync). Updates are tracked and communicated in the [OKR slides](#okr-slides).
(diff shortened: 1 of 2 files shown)


content/product-engineering/planning-process.md
```diff
@@ -70,39 +70,38 @@ There are a few ways we track plans and celebrate progress at Sourcegraph.
 
 The [GitHub project tracker](https://github.com/orgs/sourcegraph/projects/214) is **the source of truth** for key results in product and engineering. There are a few tabs.
 
-- The **Department KRs** tab tracks the top level product and engineering key results that the VPs and Directors review every week at the [VP team sync](#vp-team-sync).
-- The **Org KRs** tab tracks all the org level key results (excluding those that are tracked at the department level). VPs and Directors review this monthly, or as-needed when something is escalated by a Director.
+- The **Department KRs** tab tracks the top level product and engineering key results that the VPs and Directors review every week at the [VP team sync](#vp-team-sync). Updates are tracked and communicated in the [OKR slides](#okr-slides).
+- The **Org KRs** tab tracks all the org level key results. VPs and Directors review this monthly, or as-needed when something is escalated by a Director.
+
+  - For each KR, the assignee is responsible for posting an update to the issue at the end of each week. The update should include:
```
jasonyavorsky: What's the value of mandating this each week vs. asking people to make updates in this format when they are available? Is every KRs expected to generate wins, risks, and next steps during every week of the quarter? Is it that we are worried people will forget to give updates if we don't make them weekly and mandatory?

nicksnyder: As discussed in the sync today, going to remove this expectation from the docs. Thanks for the feedback!

PR #997, opened 2021-11-22
Clarify definitions

jasonyavorsky:


content/product-engineering/engineering/code-graph/batch-changes/metrics.md
```diff
@@ -14,5 +14,6 @@ We track success metrics in a (private) [Looker dashboard](https://sourcegraph.l
 
 ## Definitions
 
-- **Batch Changes MAU:** A visitor who triggers any event happening on a Sourcegraph Batch Changes property in a given month. In practice, we do not track CLI events, so this effectively only includes events in the GUI, or uploading a spec.
-- **Batch Changes monthly contributor:** A user that has taken an action to create or manage a batch change in a given month. In practice, a user that has previewed, or applied or closed a batch change in a given month. New actions that are added in the future, such as [comment, merge, publish](https://sourcegraph.productboard.com/roadmap/2263724-batch-changes-releases/features/6775792/portal) will be added.
+- **Sourcegraph Batch Changes Property:** TBD
```
jasonyavorsky, line 15: Not sure what to put here, but we use the term in the next sentence so should explain it.

content/product-engineering/engineering/code-graph/batch-changes/metrics.md
```diff
@@ -14,5 +14,6 @@ We track success metrics in a (private) [Looker dashboard](https://sourcegraph.l
...
+- **Batch Changes MAU:** A [visitor](../../../../bizops/user_definitions.md#visitor) who triggers any GUI event happening on a Sourcegraph Batch Changes property in a given month. We do not yet track CLI events.
+- **Batch Changes monthly contributor:** An [active user](../../../../bizops/user_definitions.md#active-user) that has previewed, applied, or closed a batch change in a given month. New actions that are added in the future, such as [comment, merge, publish](https://sourcegraph.productboard.com/roadmap/2263724-batch-changes-releases/features/6775792/portal).
```
jasonyavorsky: - Is active user correct? Or do we mean registered user?
- Is the difference between this one and the above one just visitor vs. active user? If not, what are the different events that go with each one? Someone who is a user but not a contributor doesn't immediately make sense to me, and it would help to write it down. Maybe defining what a Sourcegraph Batch Changes Property is will make it click.

malomarrec: A visitor used to be called a user, that was a rule set by BizOps a while ago. It
s very costly to update our telemetry systems away from that.

jasonyavorsky: @ebrodymoore what do you think is the right thing to do here?

ebrodymoore: Malo is right - that definition of an active user has been used since the start of on-prem @ Sourecgraph as far as I know AND it would be costly to move away from. But I also think visitors are more of a Cloud term right now since we don't distinguish most of the user definitions on-prem like we do in Cloud

jasonyavorsky: @ebrodymoore in terms of documenting the difference or making this clear/match the definitions on the analytics definition page, what would you recommend? Is there some way we can disambiguate the usage?

PR #1007, opened 2021-11-22
Update afk and ooo header

northyg:


virginiaulrich (approved):
Looks good; thanks for making the update!

PR #1008, opened 2021-11-22
Update Batch Changes strategy

malomarrec:
@jyavorska left some feedback on the Batch Changes strategy page, most of which where on conciseness.
It's much easier to add than remove, but here is my stab at it.
My goals are twofold:
(1) Making this easier to consume for readers
(2) Maintaining the usefulness of this page as our source of truth for the team

So if something was so important that after removing it from here, I would then create a google doc to add it and share it with the team, I traded conciseness off for completeness.
On the other hand, if something was or an implementation detail already represented somewhere else in an adequate way, I removed it.

I also shortened the timeframes for themes, to represent our actual focus in the next 12 months. Cloud is not that likely in that timeframe (unless we change the company strategy), and adding cloud is already mentioned in our longer-term strategy


content/company/strategy/code-graph/batch-changes/index.md
```diff
@@ -14,29 +14,19 @@ Batch Changes is a tool to find code that needs to be changed and change it at s
 
 ## Vision
 
-### 1-year vision
```
courier-new, line 17: Perhaps instead of completely omitting this section, you can leave it in but only link to "Themes" down below? It just feels like I missed something when the first time frame I see on this page is on the scale of 3 years haha.

content/company/strategy/code-graph/batch-changes/index.md
```diff
@@ -14,29 +14,19 @@ Batch Changes is a tool to find code that needs to be changed and change it at s
...
 
 ## Mission & Strategy
 
 ### Mission
 
-We allow developers to focus on changing their code, without having to do any plumbing (building scripts to interact with codehosts and integrate with other tools and workflow). We help them drive changesets to completion without having to rely on spreadsheets and painful coordination. Later on, we want to make it easier to write and run code that changes code.
+We allow developers to focus on changing their code, without having to do any plumbing (building scripts to interact with codehosts and integrate with other tools and workflow). We help them drive changesets to completion without having to rely on spreadsheets and heavy project management, as these steps are painful and time consuming, which encourages keeping old code around. We do this because we think developer's time is better spent shipping products than building plumbing and doing project management, and because having a frictionless way to make code changes will encourage all developers to keep the codebase quality high over time.
```
courier-new, line 29: 👏👏👏

courier-new (approved):

chrispine (approved):
Love it!

eseliger (approved):

jasonyavorsky (approved):
Great improvement!

mrnugget (approved):

PR #1017, opened 2021-11-23
It's 6 stars, not 8

lguychard:


eseliger (approved):
Scaling search can be so easy!

PR #1019, opened 2021-11-23
fix a confusing sentence

camdencheek:
This just adds a word that seemed to be missing that caused me some
confusion the first time I skimmed through it.


lguychard (approved):
Thanks!

PR #1023, opened 2021-11-23
playbooks/ci: update instructions on how to SSH into an agent

bobheadxi:
You can get the pod from the Buildkite build page too, if you want to see your job live

The currently documented command has a deprecation notice - updates the command to the new recommendation. We also have a default container set up, so there is no need to select (last step)


jhchabran (approved):
Thanks 🙏 !

PR #1032, opened 2021-11-23
devx: prepare newsletter in handbook, add archive

bobheadxi:
Google docs _really sucks_ with formatting and getting things ready to send. Why won't the images copy and paste????

I propose moving the entire newsletter authorship process into the handbook. If we depend entirely on `dx-announce` and review comments, we should still get a decent (possibly better than google docs, because it's more visible) collaborative process around preparing the newsletter. On top of that, we get a nice archive of each issue, _and_ it looks nice, **_and_** it is so easy to copy paste into an email.

This will deprecate the process in https://docs.google.com/document/d/1O5iUZ3cQ4c7sGhlFHDW2MXpfgLzIPFszN1jgXFUKNRo/edit#

---

image

Pastes nicely into gmail:

image


taylorsperry (approved):
Love this. +1

jhchabran (approved):

PR #1056, opened 2021-11-25
Update how to work with us docs

andreeleuterio:
Minor update to these docs since they were already accurate.


deflncha (approved):
Looks good.

PR #1077, opened 2021-11-29
Add saturation definition to metrics page

Joelkw:
To define feature saturation. Here's [what I'm trying to capture.](https://sourcegraph.slack.com/archives/CN4FC7XT4/p1638205183131800?thread_ts=1637057998.095600&cid=CN4FC7XT4)


jasonyavorsky (approved):

ebrodymoore:
@Joelkw is the denominator active user or any user with an account?

Joelkw:
@ebrodymoore that's up to individual choice – sometimes you/I want to see "of how many user accounts, how many of them have done X" and sometimes you care about "only those active this week/month." The goal was just to make it clear that "saturation" = "# feature users / a cohort of 'total' users". Should I document that it can be either, or is that not necessary?

ebrodymoore:
Got it, I think it's fine as-is! Thanks for clarifying 😄

PR #1086, opened 2021-11-30
Poker-planning added

david-sandy:
Added poker planning to security page and made small correction to title in Checkov page.


andreeleuterio (approved):
Thanks David!

PR #1108, opened 2021-11-30
Add section about #dev- channels

daxmc99:
I noticed that we don't document our `#dev-` slack channels anywhere. This adds a small blurb on them. 

The `#dev-` prefix aids in discoverability.

### DIFF
--- a/content/product-engineering/engineering/index.md
+++ b/content/product-engineering/engineering/index.md
@@ -95,6 +95,7 @@ The current channels are:
 - #dev-frontend
 - #dev-databases
 - #dev-urandom
+- #prod-eng-announce 
 
 ## Misc.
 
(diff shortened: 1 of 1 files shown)


content/product-engineering/engineering/index.md
```diff
@@ -83,6 +83,19 @@ Sourcegraph has a lot of repositories!
...
+
+The current channels are:
+
+- #dev-chat
+- #dev-ops
+- #dev-experience
+- #dev-frontend
+- #dev-databases
+- #dev-urandom
+
```
Joelkw, line 99:
```suggestion
- #prod-eng-announce 

```

@daxmc99 do you think this belongs here? I know it throws off the nice prefixing. [Until like two weeks ago](https://sourcegraph.slack.com/archives/C0EPTDE9L/p1637194172100600) this was called #dev-announce, potentially we could see if we want to change it to something like #dev-product-announce?

daxmc99: I would say it does!

Joelkw (approved):
🙌

PR #1141, opened 2021-12-01
Sourcegraphers use case for Batch Changes

malomarrec:
From time to time, we hear about Sourcegrapher/Dogfood use cases in the #batch-changes channel. There's no great source of truth for storing this, but I think this is valuable data for us and for potential teammates that are considering using Batch Changes.
I don't know where this should live, added here by default but open to suggestions.


chrispine (approved):
LGTM, thanks!

PR #1180, opened 2021-12-01
Document Batch Changes support process, update some messaging

courier-new:
We didn't have our support rotation process documented in our team processes, including the new support issues board we recently introduced, so I added a blurb to describe that. I also updated some of the messaging around communicating with our team, particularly about using @batcher-support, and removed two stated invariants that we are no longer following.


LawnGnome (approved):

eseliger (approved):

courier-new:
Noticed two other unrelated things on our doc, too. 😬 Gotta fix 'em all!!!

PR #1188, opened 2021-12-02
Introduce search product social

lguychard:
As the team has grown, time has become precious in our Monday syncs, and we rarely take the time for non-work focused conversations (eg. do tidbits).

I propose introducing "Search Product social" meetings, at EMEA + AMER-friendly times, to be more intentional about having these casual conversations. They're important! 🙂


benvenker (approved):
Good idea!

PR #1276, opened 2021-12-07
Update managed instance creation docs for executor instance filtering

kevinwojo:
Since we now have executor instances sitting alongside the application hosts, we'll need to filter out those instances for this terminal foo to work.


caugustus-sourcegraph (approved):
I didn't even think about this part breaking! I was focused entirely on "we'll have to set versions on the executors".

PR #1289, opened 2021-12-08
New employment verification process

CecilyBlack:


mercadon (approved):
So excited for this to go live! A couple of suggestions:

- For this example: "(for example, while applying for a mortgage or another type of loan)," – I think it's worth mentioning other common use cases for non-US folks (sponsorship, travel) like you had mentioned in your comms draft!
- For this point: "Direct the requestor to truework.com (or to Sourgraph's landing page)" – we can link directly to those pages for easy navigation.
- I fully trust that Truework and these instructions will cut out almost all manual requests, but I think it's worth including something in case folks have trouble or any questions, like "if you have any questions about Truework, you can reach out to People Ops (include matrix, slack channel, and/or email)

PR #1332, opened 2021-12-10
Add improved user email query

northyg:
I found this improved query in another doc and found it useful. It also returns if they have other emails and if that email is verified.


emchap (approved):
Awesome, thanks!

PR #1334, opened 2021-12-11
Clarify remove user steps for sourcegraph.com

northyg:
Added that we should inform the customer that we will remove the user name associated with the email they wrote in to us from. We may want to add that they should write back and confirm they agree to that before deleting. Something like:

'We will remove the user associated with the email you contacted us from which is <thisemail@email.com>
By removing your user and email from Sourcegraph.com, all associated data with that user will be permanently deleted. If you are okay with this, please reply back that you approve this action'

This is to be more security minded and avoid potential customer confusion. For instance, if someone writes in saying "remove user xyz", but instead we remove the email they wrote in on (or worse skip a step and remove user xyz without verifying the email in the ticket does not match.)

Additionally, I need feedback on two things:

A) This step, the template is no longer there: Create an issue in the sourcegraph/customer repository and follow the "Customer support: Delete user data" checklist issue template.

B) Related to A), The doc is confusing because would we fill out a template each time or only if it is a GDPR request? In other words, how do we differentiate between 'remove me from sourcegraph.com' and 'GDPR request, delete all my data' ?


virginiaulrich:
This looks good to me, but I am not someone who has done the process. I think @DaedalusG and/or @airamare01 would be better reviewers! I also didn't write this page, so really don't have any thoughts other than seems clear and I don't see any typos :)

content/departments/support/process/permanently_deleting_user_data.md
```diff
@@ -18,13 +18,15 @@ This will involve coordinating with multiple teams to get all the data deleted f
 
 ### In the product: how to nuke a user account
 
-If the request is for a Sourcegraph.com account, engage [one of your teammates who has access](../tools/index.md#sourcegraphcom-admin-access) (it requires site admin).
+1a. If the request is for a Sourcegraph.com account, engage [one of your teammates who has access](../tools/index.md#sourcegraphcom-admin-access) (it requires site admin). Inform the customer that you can only remove the username associated with the email they submitted the request from.
```
DaedalusG, line 21: This looks like a fine addition to me, you might consider adding a caveat for the hypothetical scenario where a person requests an account to be deactivated from an email that does not match the user email, in that scenario we should not reveal the user email to the "nuke" requestor. As this would be a way for malicious actors to gain information about Sourcegraph.com users.

DaedalusG (approved):
Added one bit of note about security concerns when communicating with requestors, but aside from that caveat looks good to me.

@virginiaulrich This doc was actually first written by Stephen

PR #1365, opened 2021-12-13
Fix severity link

nicksnyder:
This text looks like it is supposed to link somewhere but does not. Where should this link?

@andreeleuterio or @dcomas please take ownership of this PR and merge a fix.


andreeleuterio:
@nicksnyder I believe it was linking to the section below in the doc. This Incident Response doc seems due for a review, for example we have outdated severity levels. @bcreager do you know if this is in scope for the Operational Maturity work?

nicksnyder:
just looking for a quick fix now, thanks!

PR #1379, opened 2021-12-14
Fix logic for detecting redirect cycles

pdubroy:
I noticed that some of our job posting had links to the coding exercises, which were 404ing ([example](https://handbook.sourcegraph.com/engineering/hiring/software-engineer-code-review-exercise) from [Frontend Engineer - Frontend Platform](https://boards.greenhouse.io/sourcegraph91/jobs/4079706004)).

The existing logic, which was intended to catch redirect loops, was destroying valid redirect chains, by throwing away all but the last link. E.g., if there was a redirect `b -> c`, and then `a -> b` further down the list, only `b -> c` would be kept.

This PR:
- implements collapsing of redirect chains, so that all nodes in the chain always redirect to the current page
- fixes the bug, and only discards links that lead to a true cycle
- adds a test for the above ^^^
- adds jest as a dependency, and changes the `test` script to actually run the test(s).

You can see the difference between the generated `_redirects` file [here](https://github.com/sourcegraph/handbook/pull/1379/commits/c7db9c0ee4a404a26280301fff214d43f2c0f082). (I added some dummy commits to make it easy to see the diff.)


pdubroy:
You can see the difference between the generated `_redirects` file [here](https://github.com/sourcegraph/handbook/pull/1379/commits/c7db9c0ee4a404a26280301fff214d43f2c0f082). (I added some dummy commits to make it easy to see the diff here.)

jeanduplessis (approved):

jeanduplessis:
Really nice work on this @pdubroy 🙌

src/scripts/redirects.mjs
```diff
@@ -1,18 +1,43 @@
...
 /**
  * Returns all redirects that should be generated.
  *
  * @returns {Promise<{ source: string, destination: string }[]>}
  */
 export default async function redirects() {
+    const movedPages = await getMovedPagesFromHistory()
     return [
-        ...await getMovedPagesFromHistory(),
+        ...cleanupRedirects(movedPages),
```
bobheadxi: Should we call cleanupRedirects on the entire array, instead of just movedPages? That way we can catch instances of a custom redirect causing a redirect loop as well

pdubroy: Done.

package.json
```diff
@@ -122,8 +122,10 @@
     "@types/react": "17.0.37",
     "@types/relateurl": "0.2.29",
     "ajv-cli": "^5.0.0",
+    "ava": "^3.15.0",
```
felixfbecker: I think ava is not used anywhere?
```suggestion
```

src/scripts/redirects.mjs
```diff
@@ -1,18 +1,43 @@
...
+        if (nextRedirect !== source) {
+            const finalDestination = nextRedirect || destination
+            redirects.set(source, finalDestination)
+        }
+        // If nextRedirect === source, it's a direct loop — drop it entirely.
+    }
+    return [...redirects.entries()].map(([source, destination]) => ({ source, destination }))
+}
+
+export const cleanupRedirectsForTesting = cleanupRedirects
```
felixfbecker: Any reason to not just export `cleanupRedirects`?

felixfbecker (approved):
This is awesome! Redirects are such a hairball, thank you so much 🙌🏻

package.json
pdubroy: Fixed.

src/scripts/redirects.mjs
pdubroy: Good point. No, there isn't — fixed.

PR #1442, opened 2021-12-16
Fix links in deployment process for sourcegraph.com

filiphaftek:
Reason:
- when `.` is used in link description i.e. `[Deploying to sourcegraph.com]` and `-` is used in a relative link i.e. `#deploying-to-sourcegraph-com` the links are not resolved.
Right side menu works properly although.


danieldides (approved):
Wow, this is so interesting. I didn't know about this behavior with anchor links!

filiphaftek:
> Wow, this is so interesting. I didn't know about this behavior with anchor links!

@danieldides me neither before yesterday ;)

PR #1492, opened 2021-12-21
Add "affinity groups" to private Slack groups

dadlerj:


dadlerj:
Approved in https://github.com/sourcegraph/about/pull/4726, but never merged

PR #1580, opened 2021-12-30
Point at new org repo

DaedalusG:
I've moved the origin repo for cse-aws into a Sourcegraph org repo. The origin and upstream are correctly set inside the EC2 instance repo, and you should be able to clone this one and work straight from here. It's stock except for the caddy file. In the process of breaking things I may have messed up the gitconfig. So that may be a task for another day. I now have to provide a username and password (token) for most remote actions. 🤷  (ask me about my S3 debacle later lol)


benjaminwgordon (approved):
Works for me. This makes it easier to find/manage this repo for the whole team

PR #1667, opened 2022-01-05
update cloud team changes from the end of the year

bcreager:
updates for team merges and moves


jeanduplessis (approved):

nicksnyder (approved):
FYI, looks like managers have been updated in Bamboo, but not necessarily the org chart for Kevin and Crystal

PR #1717, opened 2022-01-07
Spelling and grammar updates

DocAdam:
an approved to and approved
removed repeated word: desired


DocAdam:
Combining into patch-108

PR #1718, opened 2022-01-07
Spelling update

DocAdam:
tor project management - for project management


DocAdam:
Combining into patch-108

PR #1722, opened 2022-01-07
Spelling and grammar updates

DocAdam:
explictly to explicitly 
activity around - activities around 
blog post to blog posts


DocAdam:
Combining into patch-108

PR #1727, opened 2022-01-07
Spelling and grammar updates

DocAdam:
study offers provide - study offers to provide 
Javascript - JavaScript


DocAdam:
combining into patch-108

PR #1738, opened 2022-01-10
prettier workflow: Use default GITHUB_TOKEN

pdubroy:
Switch the prettier workflow to use the default GITHUB_TOKEN (automatically available to all workflows) rather than a repository-specific token.

This allows the script to be copied as-is and reused in other Sourcegraph repos. With the current version, that's not possible because it relies on a [repository-level secret](https://github.com/sourcegraph/handbook/settings/secrets/actions), which means that you'd have to add a secret with the same name to every repository where you want to reuse this.


felixfbecker:
We need to use a different token, see context here: https://github.com/sourcegraph/handbook/pull/71

PR #1752, opened 2022-01-10
update Impact reviews note

cassiemelani:


cassiemelani:
hi @DebbieSG @mercadon @jonathanayers - requesting your review for verbiage which should have explicitly been put on our comp page back when Impact reviews were created. 

This is relevant per a conversation with @amenne, so adding her review here as well.

mercadon (approved):

PR #1757, opened 2022-01-11
Configure new managed instances to allow 24 hours for password reset emails

caugustus-sourcegraph:
👍  from Security and CE, https://sourcegraph.slack.com/archives/C1JH2BEHZ/p1641573094019500 and https://sourcegraph.slack.com/archives/CU93UDUBV/p1641838376015600


kevinwojo (commented):
👍

PR #895, opened 2021-11-16
Auto-generate product org chart

jasonyavorsky:
This uses a generator to create the product org instead of having it be manually maintained on the page, using the same data that's used for the team page (so if you update things in one place it's reflected in both, and anywhere else this content is shown).

This is also a kind of proof of concept for having this data for the whole company, which we could use for example to replace https://handbook.sourcegraph.com/company/team/org_chart.

- Before: https://handbook.sourcegraph.com/product-engineering/product/product_org
- After: https://deploy-preview-895--sourcegraph-handbook.netlify.app/product-engineering/product/product_org

Instructions: https://deploy-preview-895--sourcegraph-handbook.netlify.app/handbook/editing/add-yourself-to-team-page#additional-steps-for-product

PR #1763, opened 2022-01-11
Limit content on the org page

### LINKED ISSUE #818: Org chart page pulling in more content than desired
The [org chart](https://handbook.sourcegraph.com/company/team/org_chart) page seems to be pulling in more from other pages than intended.

In the [org chart file](https://github.com/sourcegraph/handbook/edit/main/content/company/team/org_chart.md) specific sections from other team pages are linked, indicating we might only want to show the team member info from those pages. Instead, it looks like all of most of the pages are being pulled in. This is making the page less effective, and it's breaking some links in the side bar.

For example, clicking the "Marketing" section in the right side bar brings you to the "Marketing" heading underneath the customer support section, rather than the Marketing section itself.

katjuell:
Resolves #818 by limiting the amount of content that gets pulled in on the org page. Builds on #895 by extending the idea of autogenerated content from `team.yml` to other teams.

### To Test
- Check out the branch, and head over to `team/org_chart`. Ensure that only the team members are listed and no other page information is displayed.
- Check that the updated directions and links to edit files are up-to-date and make sense.


content/team/org_chart.md
```diff
@@ -2,44 +2,68 @@
...
 
+{{generator:reporting_structure.director_legal}}
+
 ## [Tech Ops](../departments/tech-ops/index.md#members)
 
+{{generator:reporting_structure.tech_ops_manager}}
+
 ## Sales
 
 <!-- When updating the engineering team list below, please also update handbook/index.md. -->
```
katyhoward, line 49: Not sure if we still need this note considering the updated directions?

data/team.yml
```diff
@@ -2276,9 +2501,103 @@ philipp_spiess:
...
   role: Full-Stack Engineer, Growth and Integrations
+  reports_to: engineering_manager_growth
   location: Vienna, AT 🇦🇹
   pronunciation: '[ˈʃpiːs](http://ipa-reader.xyz/?text=%CB%88%CA%83pi%CB%90s&voice=Marlene)'
   email: philipp.spiess@sourcegraph.com
   github: philipp-spiess
   links: '[@PhilippSpiess](https://twitter.com/PhilippSpiess), [LinkedIn](https://www.linkedin.com/in/philipp-spiess/)'
   description: Philipp grew up in the countryside of Austria and developed a passion for Web programming very early. After moving to Vienna to study computer science and working on some freelancing gigs, he helped create a PDF editor for the Web. Before joining Sourcegraph, Philipp worked as a Front-End Engineer at Meta/Facebook in London. Outside of work, his hobbies vary wildly, from being a home barista to learning new cooking techniques, working on DIY projects, or playing video games.
+
+jessie_char:
```
katyhoward: Wondering if it would be a good call to get someone from the people team to give these added names a quick look over, once we're good with everything else.

jasonyavorsky (approved):
Amazing work @katjuell! This looks great. I think there are two parts that need to happen soon after this merges:

- Some kind of general communication on how to maintain the data. Perhaps HR has some role in it.
- Replace existing org charts on other pages with the generator text so there's a single source of truth. For example, https://handbook.sourcegraph.com/departments/sales#members has a manually generated list that should be replaced with the autogenerated one.

Some teams may be doing slightly unique things with their org charts, like having different links or structures, and they need to be ok with losing those.

content/team/org_chart.md
jasonyavorsky: I think not needed, once you make sure that all the manually created org charts in other places in the handbook (that this original page was pulling from) are replaced with generators.

katyhoward: Awesome! Thanks for the callout about the other org charts.

katyhoward:
Thanks for this feedback @jyavorska ! After talking with @marybelzer this morning, we decided that we'll do an information blast in `handbook-announce` to let folks know that the change is live, and to remind them to add themselves in `team.yml` if they don't see themselves in the org chart. We opted to take folks out who were not already in there and who were not managers, just to give them a chance to opt-in.

For 2, I've created a follow-up issue #1771 to handle the other charts on other team pages. Thanks again for the callout!

Last thing — I do not see the avatar issue from [this comment](https://sourcegraph.slack.com/archives/CQ44Y7F4G/p1641998346019500) since I don't see avatars on the org page. I'll keep an eye out elsewhere for it though, but figure it's safe to say it's not part of the current work on this page.

PR #1793, opened 2022-01-13
Add org charts to team pages.

### LINKED ISSUE #1771: Autogenerate org charts across team pages

katjuell:
Resolves #1771 by adding org charts to team pages. Uses a combination of the existing `generateReportingStructure` function, currently used on our main org chart, and a new function, `generateTeamOrgChart` that can accept an argument and reads data from the `product_teams.yml` file. Teams that currently have custom code on their team page, including avatars, were not updated in this PR.


content/handbook/editing/add-yourself-to-team-page.md
```diff
@@ -43,10 +43,13 @@ director_design:
...
   manager_role_slug: director_design
-  hidden_on_team_page: true
 ```
 
-Remove `(Interim)` from the role title, as well as the `hidden_on_team_page` field. You don't need to change the `manager_role_slug` value.
+Remove `(Interim)` from the role title. You don't need to change the `manager_role_slug` value.
+
+### Product team pages
+
+Most of our individual product team pages have org charts that are autogenerated from the [product_teams.yml](https://github.com/sourcegraph/handbook/blob/main/data/product_teams.yml) file. If you are a new manager joining a team, please make sure to check this file, and to add yourself there as well.
```
katyhoward, line 52: Not sure if there is another place to document this? I updated the instructions for managers on the org chart as well, but are these the only places where we would need instructions?

schema/team.schema.json
```diff
@@ -35,9 +35,6 @@
       },
       "description": {
         "type": "string"
-      },
-      "hide_on_team_page": {
```
katyhoward, line 39: It wasn't clear to me that we were hiding interim folks on the org chart or on team pages? But please correct me if I'm wrong. I was looking at [this page](https://handbook.sourcegraph.com/departments/product-engineering/product/team) for reference.

chrispine:
Currently, our page lists the entire Batch Changes team, including our PM (Malo) and Designer (Rob). Because they don't report to me (the EM), does that mean they will not be listed as being on the team? Because they are definitely part of our team! 😃 

Edit: Now that I read further, it looks like it's handled correctly? But I'm not sure, so just wanted to call it out. Thanks!

unknwon (approved):

katyhoward:
Hi @chrispine — thank you for calling this out! I think everyone should be represented on the `batch-changes` page, but just for a sanity check, here's what I see on the [branch deploy](https://deploy-preview-1793--sourcegraph-handbook.netlify.app/departments/product-engineering/engineering/code-graph/batch-changes):
Screen Shot 2022-01-14 at 1 40 36 PM

And then here's what I see in prod:
Screen Shot 2022-01-14 at 1 40 26 PM

Please let me know if that doesn't look right to you!

PR #1825, opened 2022-01-17
Async improvements for working groups

jasonyavorsky:
This adds some ideas to working groups to help them work better in an async way by doing things like writing things down, recording meetings, and so on.


jasonyavorsky:
@christinaforney anything else you'd add? I think this is non-controversial and in line with company values so I think I'll just merge if you think it makes sense.

christinaforney (approved):
lgtm

PR #1851, opened 2022-01-18
Single copy of use cases

jasonyavorsky:
This makes it so there's a single copy of the list of use cases on this page, and one place links to the other. This will help ensure a single source of truth so we don't accidentally update one copy and not the other.


content/strategy-goals/strategy/index.md
```diff
@@ -83,26 +83,13 @@ Let's break that down:
 - [Iterate with existing customers first](#iterate-with-existing-customers-first)
 - [Grow ARR](#grow-arr)
 
-### Target 5 specific use cases
+### Use cases
```
jasonyavorsky:
```suggestion
### Target 5 specific use cases
```

PR #1892, opened 2022-01-20
Add terminology to glossary to avoid overloading "onboarding"

quinnkeast:


jasonyavorsky:
This looks good to me but probably needs more eyes on it before we can make it official.

jasonyavorsky (approved):

jasonyavorsky:
@quinnkeast given this effects terms that almost everyone at the company uses, consider sharing in #general

PR #1953, opened 2022-01-25
Update expense instruction for non-us teammates

michaellzc:
Per [this thread](https://sourcegraph.slack.com/archives/C02FSM7DU/p1642545186112500), non-us teammates have the option to order a company card for expenses to avoid tax implications.


desenesterling (changes requested):
as mentioned twice (email and slack) - Accounting will be updating this page when the policy is live.

PR #1974, opened 2022-01-26
Adding Security as one of the approvers for new software

dcomas:
Adding Security as the approver for new software


dcomas:
Merging after 👍  from Tommy O'Donnell

PR #1996, opened 2022-01-27
Add link to use cases mapping

Joelkw:
We might long-term want to standardize this kind of quick reference for each product team?


jasonyavorsky (approved):
Good add, I also added a link from the use case index.

jasonyavorsky:
Probably in the end each use case summary page should link back to the feature areas that enable it.

PR #2001, opened 2022-01-27
Add feature lists for each use case

jasonyavorsky:
This adds feature lists to each use case page via data and a generator. It also allows going the other direction (from the product team strategy to the use cases.)


data/features.yml
```diff
@@ -5,6 +5,8 @@ literal_search:
   product_team: search_product
   maturity: ga
   documentation_link: https://docs.sourcegraph.com/code_search/reference/queries
+  use_cases:
+    - developer_onboarding
```
jasonyavorsky, line 13:
```suggestion
    - none
```

schema/use_cases.schema.json
```diff
@@ -0,0 +1,16 @@
...
+          "type": "string"
+        },
+        "link": {
+          "type": "string"
+        }
+      },
+      "required": ["title", "link"]
+    }
+  }
+
```
jasonyavorsky:
```suggestion

```

schema/use_cases.schema.json
```diff
@@ -0,0 +1,15 @@
...
+        "title": {
+          "type": "string"
+        },
+        "link": {
+          "type": "string"
+        }
+      },
+      "required": ["title", "link"]
+    }
+  }
```
jasonyavorsky:
```suggestion
  }

```

data/features.yml
```diff
@@ -291,6 +327,8 @@ code_insights:
   product_team: code_insights
   maturity: beta
   documentation_link: https://docs.sourcegraph.com/
+  use_cases:
+    - none
```
Joelkw:
```suggestion
    - code_health
    - fixing_security_vulnerabilities
    - incident_response
    - code_reuse
    - developer_onboarding
```

Joelkw (approved):

jasonyavorsky:
@benvenker @lguychard @macraig I would have thought you'd have had some features related to use cases - can you confirm you've checked and don't before I merge?

data/features.yml
```diff
@@ -291,6 +327,12 @@ code_insights:
   product_team: code_insights
   maturity: beta
   documentation_link: https://docs.sourcegraph.com/
+  use_cases:
```
jasonyavorsky, line 367: @Joelkw I saw you left a suggestion with this same content, was that to avoid these appearing due to the embargo?

data/features.yml
jasonyavorsky: Resolving since this was applied in a commit

data/features.yml
Joelkw: @jyavorska thanks for checking! No, I think this is so high level that it's fine, since the use cases don't tell you anything about the feature, what it does, how it works, etc. That was probably me missing the final merge button, thought I merged them in, sorry.

lguychard:
@jyavorska added use cases for Search features.

jasonyavorsky:
@dan-mckean there are some features in here owned by repo management & delivery; are there any that you'd want to associate with any of the use cases?

data/features.yml
```diff
@@ -371,6 +443,8 @@ browser_extension:
   product_team: repo_management
   maturity: ga
   documentation_link: https://docs.sourcegraph.com/integration/browser_extension
```
dan-mckean, line 452: We don't own that, does the Growth team maybe?

data/features.yml
```diff
@@ -352,6 +422,8 @@ user_authentication:
   product_team: repo_management
```
dan-mckean: This is actually Cloud SaaS who owns User auth - though I'm not sure exactly how to name them in this file to avoid breaking things.

dan-mckean (commented):
@jyavorska I've added a couple of comments. But I don't think there's any use cases that we impact directly enough to warrant listing against the things for either Repo Mgmt or Delivery.

data/features.yml
jasonyavorsky:
```suggestion
  product_team: cloud_saas
```

This should do the trick. We'll get an error message if I get it wrong.

data/features.yml
jasonyavorsky: Ok, thanks!

data/features.yml
```diff
@@ -371,6 +443,8 @@ browser_extension:
...
```
jasonyavorsky:
```suggestion
  product_team: growth_and_integrations
```

PR #2012, opened 2022-01-27
Clarify license key generation process

simonintech:
Update description of the license key generation process to ensure the following:
1. format of username associated with subscription
2. accounts should not have an associated email address
3. license keys should be sent directly to customer


content/departments/ce/process/license_keys.md
```diff
@@ -6,12 +6,12 @@ Valid license keys can only be generated by site administrators on Sourcegraph.c
 
 License keys are created and managed on Sourcegraph.com in the **Site-admin > Subscriptions** page. Once a license key has been created for a user, they can sign in to Sourcegraph.com to access it from their user profile.
```
emchap:
```suggestion
License keys are created and managed on Sourcegraph.com in the **Site-admin > Subscriptions** page. 
```

PR #2041, opened 2022-01-31
Update Cloud teams info

anna-mmikhova:
Update cloud teams info (where available)


malomarrec (approved):

jasonyavorsky:
@anna-mmikhova this was failing because the extensions team was renamed, but there were still features that were pointed at the now missing extensions team. I updated them to point to the growth & integrations team, is that right?

jasonyavorsky:
It was also failing because Murat's name was misspelled

jasonyavorsky:
One more question @anna-mmikhova - the delivery team was deleted, but there's still a page in the handbook at strategy_link: /company/strategy/enablement/delivery/index.md. Is that intentional?

anna-mmikhova:
I left the delivery link under DevOps for now while we restructure the Cloud org pages, it's quite a mess

PR #2049, opened 2022-01-31
Resolve trailing slash behavior

### LINKED ISSUE #1822: Bad links from "Benefits, Pay, and Perks" index page
The links from the "Benefits, Pay, and Perks" [index page](https://handbook.sourcegraph.com/benefits-pay-perks/) are broken, seemingly because they add an extra `/benefits-pay-perks` to the slug. For example, the holidays link redirects to `https://handbook.sourcegraph.com/benefits-pay-perks/benefits-pay-perks/benefits-perks/holidays`. 

I looked at fixing this, but interestingly, I can't reproduce the issue locally.

### LINKED ISSUE #2307: Broken link detection does not seem to be working
At one point if you had a link to a local page that didn't work it would tell you, but this seems to have stopped working. https://handbook.sourcegraph.com/departments/tech-ops/process/ (at least as of the time of creating this issue) has a link to a local page, but it returns a 404.

katjuell:
Problem: The [default trailing slash behavior](https://nextjs.org/docs/api-reference/next.config.js/trailing-slash) that Next.js provides is not being respected in prod or on staging, though it is respected locally. What this means in practice is that if a URL contains a trailing slash, that slash will not be rewritten and will break our linking logic (which is predicated on Next's default behavior).

Setting behavior at the proxy level unfortunately does not work: setting [`pretty_urls`](https://docs.netlify.com/routing/redirects/redirect-options/#trailing-slash) to [`false`](https://github.com/sourcegraph/handbook/pull/2049/commits/0f163084a78f322c04b9c604d891ef3ebff85a20) in the Netlify config does not have the desired outcome. The discussion around Netlify and Next.js (in the absence of something like an Nginx proxy) seems to favor enforcing trailing slash behavior everywhere (at both the Next config and Netlify levels). 

This PR also resolves a similar issue with anchor links on index pages: currently, a trailing slash gets inserted when navigating to something like `/strategy-goals/strategy/#use-cases`, which subsequently breaks linking logic from that page. For example:
```
https://handbook.sourcegraph.com/departments/talent/process/recruitment-branding -> click on making referral link ->
https://handbook.sourcegraph.com/departments/talent/process/#making-a-referral ->
any subsequent link to another page
```
These cases should now be resolved.

This PR will resolve #1822 & other cases where we have seen where links that have a trailing slash produce 404s.


felixfbecker:
I think it's not a good solution to do this on the client side, because it means the client will make an HTTP request, get a 200, then JavaScript will run and change the URL to something else and trigger another HTTP request (?). Redirects should always happen on the server to make sure that clients (such as browsers, but also search engine crawlers, our search, etc) will correctly use the canonical address.

Mind going into more detail on this? (why does it not work/what happens?)

> Setting behavior at the proxy level unfortunately does not work: setting [pretty_urls](https://docs.netlify.com/routing/redirects/redirect-options/#trailing-slash) to [false](https://github.com/sourcegraph/handbook/pull/2049/commits/0f163084a78f322c04b9c604d891ef3ebff85a20) in the Netlify config does not have the desired outcome. The discussion around Netlify and Next.js (in the absence of something like an Nginx proxy) seems to favor enforcing trailing slash behavior everywhere (at both the Next config and Netlify levels). This does not seem like the best direction to go in, however, given the current logic of our app, which anticipates the Next default behavior.

katyhoward:
Thanks for this feedback! I 100% agree with you re: client side, and was at first pass hoping that Netlify would provide a simple way to respect Next's default behavior. Flipping `pretty_urls` to `false` did not solve the problems we've been seeing, though (tested it twice, just to make sure, in [`0f16308`](https://github.com/sourcegraph/handbook/pull/2049/commits/0f163084a78f322c04b9c604d891ef3ebff85a20) and [`0b2d1b5`](https://github.com/sourcegraph/handbook/pull/2049/commits/0b2d1b526dd8c382b0a1795bcb806049bef8a26c)). Trailing slashes remained, and links remained broken.

I had been concerned about the duplicate content penalty if we flipped trailing slash to true, but maybe that's not as much of a concern as I think it is? At any rate, that's where this PR is at this point: flipping the trailing slash switch on Netlify, with a small change to handle anchor links on index pages, does seem to have resolved the issues that Mary and I know about.

jeanduplessis (approved):
@katjuell I think the solution to switch to enforcing trailing slashes is the best solution for the current situation.

One thing that I would encourage you to just have a look at is whether we should update any content in the handbook where we provide instructions around linking and if there's a need to make any announcement in #handbook-announce.

jeanduplessis:
@katjuell I think this is a side-effect of the change. On this page, https://deploy-preview-2049--sourcegraph-handbook.netlify.app/benefits-pay-perks/benefits-perks/spending-company-money/#team-specific-totals, if you click on the Design team link it now results in a broken link.

The link works on the production build. Would we not need to update relative links in the code?

katyhoward:
@jeanduplessis thanks for surfacing this! I think that behavior should be resolved now. Checking links from index and non-index pages and both look okay.

jeanduplessis (approved):
@katjuell seems good now 👍

katyhoward:
For documentation, I wonder if it would make sense to leave the directions as they are? If these changes don't require any different behavior on the part of handbook contributors?

jeanduplessis:
> For documentation, I wonder if it would make sense to leave the directions as they are? If these changes don't require any different behavior on the part of handbook contributors?

@katjuell if there's no change to how people need to do things, then keeping things as is, is fine.

jasonyavorsky:
This was a tricky fix, thanks for getting it done!

PR #2136, opened 2022-02-02
Team names

coury-clark:
This PR adds a `team` field to the teammates data file that will render in the autogenerated org chart. I did some basic data clean up based on some regexp to try and move the data previously in `role` to `team`, but this was spotty at best.

This will append a suffix of `team` to anybody in the org chart whose `team` does *not* match their managers `team`. This way, teams that share a common value (if everyone puts the same tag) will only be visible at the root of that tree to reduce clutter.

I share the same team as Felix
image

I don't share the same team as Felix
image

### DIFF
--- a/src/lib/generatedMarkdown.js
+++ b/src/lib/generatedMarkdown.js
@@ -162,7 +162,7 @@ function getReports(teamMembers, role_slug, parentTeam, indent) {
   for (const [teamMemberName, teamMember] of Object.entries(teamMembers)) {
     if (teamMember.reports_to === role_slug) {
       const currentTeam = createValidTeamString(teamMember)
-      const teamString = parentTeam && parentTeam === currentTeam ? '' :currentTeam
+      const teamString = parentTeam && parentTeam === currentTeam ? '' : currentTeam
       console.log(indent)
       const spaces = ' '.repeat(indent * 2)
       content += `${spaces}- [${String(teamMember.name)}](/team/index.md#${String(
@@ -191,7 +191,7 @@ export async function generateReportingStructure(starting_role) {
       )}), ${String(teamMember.role)}\n`
     }
   }
-  pageContent += getReports(teamMembers, starting_role,'', 1)
+  pageContent += getReports(teamMembers, starting_role, '', 1)
   return pageContent
 }
 
--- a/src/lib/generatedMarkdown.js
+++ b/src/lib/generatedMarkdown.js
@@ -14,6 +14,10 @@ function createValidTeamAnchor(name) {
   return name.toLowerCase().replace(/\s+/g, '-')
 }
 
+function createValidTeamString(teamMember) {
+  return teamMember.team ? `- ${String(teamMember.team)}` : ''
+}
+
 function createBioLink(name) {
   const bioLinkUrlPrefix = '/team/index.md#'
   return `${bioLinkUrlPrefix}${String(createValidTeamAnchor(name))}`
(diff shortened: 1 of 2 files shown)


jasonyavorsky:
You might want to add these new teams to product_teams.yml (and maybe remove the product_ prefix) so that there's some consistency, and then point to those team objects instead of having a free text field for everyone to enter their team name. The nice thing about using the data references is you only have to update things in one place when they change, and there is reduced risk of typos causing weird behavior.

You could also then validate the relationships via validateData.mjs.

src/lib/generatedMarkdown.js
```diff
@@ -153,18 +157,22 @@ export async function generateCodeHostsList() {
...
 }
 
-function getReports(teamMembers, role_slug, indent) {
+function getReports(teamMembers, role_slug, parentTeam, indent) {
   let content = ''
   for (const [teamMemberName, teamMember] of Object.entries(teamMembers)) {
     if (teamMember.reports_to === role_slug) {
+      const currentTeam = createValidTeamString(teamMember)
+      const teamString = parentTeam && parentTeam === currentTeam ? '' : currentTeam
+      console.log(indent)
```
coury-clark:
```suggestion
```

jasonyavorsky:
You might also want to update the documentation for people to add themselves to the team page before merging to explain how this works.

jasonyavorsky:
One gap in the model that's implemented here is that there are three organizational levels:

- Team
- Org
- Department

Some people are members of one, two, or three of them (for example, a department head would only be a member of the department and not one/all of the individual orgs within the department, and the same goes for the individual teams that make up the orgs).

Maybe there's a different data structure that could represent that kind of membership?

coury-clark:
@jyavorska Yeah, it's definitely not a perfect solution. It's complicated because technically this is a many-many relationship, which are difficult to display without multiple ways to slice the views. One thing to note is that this isn't trying to represent total set membership - that is for teammate X what are all of the sets (teams) they belong to; instead, it is trying to provide a simple mechanism to answer the question "who is on team X".

The reason I thought maybe this wasn't such a problem was the way I implemented the parent / child override. The idea is that in the leadership levels you may have many different team names, but as you get into the IC levels most members would share a team under whoever they report to. That probably isn't always true, but seems true enough at the moment.

> You might also want to update the documentation for people to add themselves to the team page before merging to explain how this works.

Agreed - just wanted to get a sense of whether or not this is something that is valuable. Appreciate the feedback!

coury-clark:
> You might want to add these new teams to product_teams.yml (and maybe remove the product_ prefix) so that there's some consistency, and then point to those team objects instead of having a free text field for everyone to enter their team name. The nice thing about using the data references is you only have to update things in one place when they change, and there is reduced risk of typos causing weird behavior.
> 
> You could also then validate the relationships via validateData.mjs.

Yeah - that's a good idea. I was also thinking maybe just invert the reference, and derive the team name from the parent, but I think that probably isn't as accurate as what's in the PR.

jasonyavorsky:
@coury-clark can I help in any way with this?

coury-clark:
@jyavorska I've been pretty swamped with team obligations for Code Insights. I may have a chance to come back to this in a few weeks, but if you want you can take over this PR!

jasonyavorsky:
Sure, I'll see if I have time too. There's no rush or anything but just wanted to offer help in case it got stuck for some reason I could help with.

PR #2144, opened 2022-02-03
Remove old core app strat page

malomarrec:
As a first step, I temporarily moved the content to cloud-saas as it was related to cloud roadmap, but it's not super up to date. That way nobody will be confused about Core App.


quinnkeast (approved):
Good interim fix. Will update from here this week.

PR #2145, opened 2022-02-03
Mm/q1 update

malomarrec:
Not included: refresh the competitive analysis, which I will get to in coming weeks.


eseliger (commented):
LGTM

PR #2146, opened 2022-02-03
Use consistent terms for experimental and beta

malomarrec:


Joelkw (approved):
thanks!

quinnkeast (changes requested):
I don't think this is actually accurate—we may need another way of putting this in the handbook.

Here, it's intended to clarify "early access enrollment" as the term in place of "onboarding customers" in what we're providing to Cloud customers right now. I'd describe the difference as "experimental describes the feature itself, early access enrollment describes the customer relationship with the feature + the action we're taking.

I'm not quite sure how to solve this... any thoughts?

malomarrec:
@quinnkeast What do you think about the new proposal? (I assumed early access assumed "to a feature in beta or experimental", I don't see how we could do early access to something that is General Admission, that would be contradictory)

quinnkeast (approved):
Brilliant!

PR #2274, opened 2022-02-09
Add name spacing

rrhyne:
Adds a quick description of how to namespace event names.


ebrodymoore:
This happens [here](https://handbook.sourcegraph.com/departments/bizops/tools/amplitude#amplitude-govern) - let's keep it that way for now for consistency (we may end up changing this later)

rrhyne:
The suggested PR helps guide users in creating names that are understandable from Amplitude with little research. It would help keep amplitude from being a mess of events names: CTAButtonClicked instead of CodeMonitorCTAButtonClicked and SavedSearchCTAButtonClicked.

I am educating people manually about this now, which is why I think it deserves a place in the handbook and I don't see where this specific issue is addressed at your link.

ebrodymoore (approved):
My bad, misread the PR

PR #2304, opened 2022-02-11
Add embedded notebooks support to handbook

jasonyavorsky:
Example: https://deploy-preview-2304--sourcegraph-handbook.netlify.app/strategy-goals/strategy/use-cases/fixing-security-vulnerabilities#log4j-search-notebook

image

Issues:
- The embed itself seems to be honoring media queries but the handbook site doesn't support them, so if your computer is in dark mode you get a dark notebook against a light site which looks strange. See discussion below; fixed via https://github.com/sourcegraph/sourcegraph/pull/31054.
- The embedded iframe would benefit from including a button to open it on Sourcegraph. This is true for anywhere it's embedded, not just in the handbook.
- Clicking on links inside the `iframe` results in a Chrome error (this is being worked on)
- There's a lot of white space at the bottom of the notebook.

It's usable with the above issues, I think. Both will need to be fixed in the implementation of embeds, not here, so once that's done the handbook will automatically benefit from them. Better to dogfood the feature in the meantime, IMHO.


jasonyavorsky:
@novoselrok is there a way to ask the embedded notebook to come back always in light mode? It's kind of ugly now because it mixes it (handbook doesn't support dark mode yet).

novoselrok:
> @novoselrok is there a way to ask the embedded notebook to come back always in light mode? It's kind of ugly now because it mixes it (handbook doesn't support dark mode yet).

Right now, embedded notebooks only support light mode. In the future we can add a URL param (e.g., `?theme=dark`) to switch to dark mode, but the default will be light mode.

jasonyavorsky:
@novoselrok it's showing up as dark against light on mine. Chrome on MacOS, system set to dark mode.

image

novoselrok:
> @novoselrok it's showing up as dark against light on mine:
> 
> image

Very weird, when I run the branch locally it seems to be working fine:
Screenshot 2022-02-11 at 12 07 25

Can you try setting an explicit background color on the iframe and on the ` <div class="embed notebook">` element. I think the iframe is picking up a weird background color from somewhere :/

jasonyavorsky:
@novoselrok it's coming from the media query, you need to disregard it somewhere in the embed code. If I set my computer to light mode, then the embed also returns as light mode.

jasonyavorsky:
Cc @benvenker for the issues I noticed while getting this working (see description).

content/strategy-goals/strategy/use-cases/fixing-security-vulnerabilities.md
```diff
@@ -63,3 +63,9 @@ The following are real, anonymized quotes from customers today on how we’re al
 - [Productboard fixing security vulnerabilities (internal only)](https://sourcegraph.productboard.com/feature-board/3957049-fy23-use-cases/features/11482297/detail)
 - [Tackling Log4j with Sourcegraph](https://about.sourcegraph.com/blog/log4j-log4shell-0-day)
 - [Cloudflare case study](https://about.sourcegraph.com/case-studies/cloudflare-accelerates-debugging-and-improves-security/)
+
+### Log4j search notebook
+
+As part of our log4j mitigation efforts, we used a Search Notebook to demonstrate how to use Sourcegraph to solve this use case. You can see that use case here.
```
jasonyavorsky:
```suggestion
As part of our log4j mitigation efforts, we used a Search Notebook to demonstrate how to use Sourcegraph to solve this use case. You can see that here.
```

novoselrok:
> @novoselrok it's coming from the media query, you need to disregard it somewhere in the embed code. If I set my computer to light mode, then the embed also returns as light mode.

You're right, here is the fix in the main repo: https://github.com/sourcegraph/sourcegraph/pull/31054

jasonyavorsky:
@novoselrok another strange thing, wondering if you have it too. If I click on the blog post link inside the embed at https://deploy-preview-2304--sourcegraph-handbook.netlify.app/strategy-goals/strategy/use-cases/fixing-security-vulnerabilities#log4j-search-notebook, it doesn't work. I wonder if all links inside the embedded content need to be rewritten to target a new tab?

benvenker:
@jyavorska I'm experiencing the same thing. I get the chrome sad face in the embed when I click on the link. Tab+click works fine in a new tab though.

novoselrok:
> @novoselrok another strange thing, wondering if you have it too. If I click on the blog post link inside the embed at https://deploy-preview-2304--sourcegraph-handbook.netlify.app/strategy-goals/strategy/use-cases/fixing-security-vulnerabilities#log4j-search-notebook, it doesn't work. I wonder if all links inside the embedded content need to be rewritten to target a new tab?

Yes, that's the unfortunate consequence of using an iframe. The specific blog post link in the markdown will have to changed to use `target="_blank"`. But the issue goes further than that. Because our web app assumes it is not embedded, it tries to open everything within the current (iframe) window, rather than opening within the top (user) window. I'll investigate if we can automatically open all of our links in a new tab.

benvenker:
Ah, yeah I remember us talking about this issue with search results as
well. Would this be partially resolved if we changed that behavior in the
app?

benvenker:
@novoselrok yeah I remember us talking about this now. Would this be partially fixed if we updated our app to open results or links from Notebooks in a new tab?

novoselrok:
> @novoselrok yeah I remember us talking about this now. Would this be partially fixed if we updated our app to open results or links from Notebooks in a new tab?

Yep, partially. But I think I have a simple enough fix for the embedded web app. We would need to switch the link component we're using to the one that always opens links in a new tab (just for the embedded notebooks).

PR #2315, opened 2022-02-11
Update Triaging SLA and AER 1st Day.

shawnteeharris:


virginiaulrich (approved):
Marking this as approved and with a tiny request -- can we add to #7 ... that when a SLA is close to expiring, we ask those AERs with less than 5 open/active tickets (pending or open ... on hold = not active) if they have capacity to take it. I think that number may matter for knowing who to at mention.

shawnteeharris (commented):
Updated Missing SLA

PR #2322, opened 2022-02-14
Require `CONTEXT=production` for GitHub data fetches

jasonyavorsky:
This requires `CONTEXT` to be set to `production` for fetching commit data, which is used for the display of recent contributors. This will make it so that only production builds fetch this data, reducing the likelihood of rate limits and speeding up the build.

The current usage of `GITHUB_TOKEN` was not documented, so in addition to documenting the use of the `CONTEXT` variable in local builds, it explains that as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,10 @@ Then start the handbook:
 yarn dev
 ```
 
+### Fetching GitHub data
+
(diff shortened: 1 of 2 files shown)


jeanduplessis (approved):

jeanduplessis:
LGTM.

PR #2325, opened 2022-02-14
Force light theme for search notebooks on handbook

jasonyavorsky:
Implements handbook side fix for https://github.com/sourcegraph/sourcegraph/pull/31159


limitedmage:
This is 100% out of scope for this, but will the handbook ever support dark mode? Is there an issue tracking this?

jasonyavorsky:
@limitedmage I think I recall that being discussed at one point, I'd ask in #handbook

PR #2344, opened 2022-02-15
Brief strategy updates

Joelkw:
Minor updates given the embargo restrictions of GA at the moment. Major update is the tentative OKR; will discuss more thoroughly with the team in two weeks but don't want to distract them now. That said, it feels like the right metric for this quarter (before pivoting to feature-usage / behavior metrics next quarter) so the question is just the number for the metric.

### DIFF
--- a/content/strategy-goals/strategy/code-graph/code-insights/index.md
+++ b/content/strategy-goals/strategy/code-graph/code-insights/index.md
@@ -157,7 +157,7 @@ The next six months are focused on building a product that can quickly prove out
 
 ### Goals
 
-#### FY23Q1
+These are the goals we plan to fully complete by the end of Q1 FY23:
 
 - KR 1: 10 customers have purchased Code Insights in FY23Q1
 
--- a/content/strategy-goals/strategy/code-graph/code-insights/index.md
+++ b/content/strategy-goals/strategy/code-graph/code-insights/index.md
@@ -125,11 +125,11 @@ Users do not need to learn a new query language or format to use Code Insights. 
 
 ## Where we are now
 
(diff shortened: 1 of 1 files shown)


content/strategy-goals/strategy/code-graph/code-insights/index.md
```diff
@@ -157,9 +157,9 @@ The next six months are focused on building a product that can quickly prove out
 
 ### Goals
 
-#### FY22q4
+#### FY23Q1
```
jasonyavorsky: Minor thing but I think we can remove this heading. The reader can assume whatever is here are the current goals, and there isn't any other section for goals in other quarters.

jasonyavorsky:
Not much to add, nice brief update. Thanks!

jasonyavorsky (approved):

content/strategy-goals/strategy/code-graph/code-insights/index.md
yinquanteo: I think clarity here is helpful. If we removed it, while the reader can assume the goals are current, it isn't clear what scope "current" refers to. E.g. it could be the current quarter, but it could also be the current financial year.

jasonyavorsky: Then let's say something like "We plan to complete these goals by X." Writing down the name of the current quarter as a section heading is less clear/useful.

jasonyavorsky:
```suggestion
These are the goals we plan to fully complete by the end of Q1 FY23:
```

jasonyavorsky:
Setting to automerge since all feedback is in, hope you don't mind @Joelkw

PR #2367, opened 2022-02-16
Add deployment options matrix

malomarrec:
We are missing a page for outlining what feature is supported on self-hosted, managed instances, or cloud only.
This is a proposal to add such a page.
[Slack discussion](https://sourcegraph.slack.com/archives/C01SD4A49E1/p1643966159528069)


malomarrec:
TIL: need to add a schema.

jasonyavorsky:
https://handbook.sourcegraph.com/handbook/editing/content-generation may be helpful.

schema/deployment_options.schema.json
```diff
@@ -0,0 +1,12 @@
+{
```
jasonyavorsky, line 1: @malomarrec schema/features.schema.json should probably change too to account for the new possibilities in that file.

jasonyavorsky (approved):

schema/deployment_options.schema.json
malomarrec: ty!

malomarrec:
@Joelkw @macraig @ryphil - I am merging this, but feel free to edit the data files to represent Cloud availability. I don't think we need to be exhaustive about what _is_ supported where (that would create a huge, noisy page), rather we should highlight what is not supported.

jasonyavorsky:
Really nice change @malomarrec thanks for this. Remember to share in some channel so people are aware.

PR #2385, opened 2022-02-16
adds release guild members, and auto generated guild membership

coury-clark:
Adds a new type of membership called `guild`. This can autopopulate a list of members from team references.
CleanShot 2022-02-16 at 12 49 19@2x

### DIFF
--- a/data/guilds.yml
+++ b/data/guilds.yml
@@ -1,0 +1,4 @@
+release_guild:
+  title: Release Guild
+  leader: coury_clark
+  member_refs: [coury_clark, joe_chen, camden_cheek, tomás_senart, keegan_carruthers-smith, mohammad_umer_alam]
--- a/data/guilds.yml
+++ b/data/guilds.yml
@@ -1,4 +1,4 @@
 release_guild:
   title: Release Guild
   leader: coury_clark
-  member_refs: [coury_clark, joe_chen, camden_cheek, tomás_senart, keegan_carruthers-smith, mohammad_umer_alam]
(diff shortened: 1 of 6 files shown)


michaellzc:
so many CI checks 😂

looks like we need a `guilds.schema.json` defined for the new yaml

https://github.com/sourcegraph/handbook/runs/5222333622?check_suite_focus=true#step:6:11

coury-clark:
@michaellzc  

ah yeah - just added that

data/guilds.yml
```diff
@@ -0,0 +1,4 @@
+release_guild:
+  title: Release Guild
+  leader: coury_clark
+  member_refs: [coury_clark, joe_chen, camden_cheek, tomás_senart, keegan_carruthers-smith, mohammad_umer_alam]
```
jasonyavorsky: Maybe just call this `members`?

jasonyavorsky:
Since you're adding references, you should also update https://github.com/sourcegraph/handbook/blob/main/src/scripts/validateData.mjs to make sure that the refs you point to exist.

See https://handbook.sourcegraph.com/handbook/editing/content-generation/#code for more info.

jasonyavorsky (approved):

coury-clark:
> Since you're adding references, you should also update https://github.com/sourcegraph/handbook/blob/main/src/scripts/validateData.mjs to make sure that the refs you point to exist.
> 
> See https://handbook.sourcegraph.com/handbook/editing/content-generation/#code for more info.

Added!

jasonyavorsky:
Nice job @coury-clark !

PR #2397, opened 2022-02-17
Add support for hiding the sidebar on certain pages

### LINKED ISSUE #1934: Product compatibility table overflows onto right-side menu
On [this page](https://handbook.sourcegraph.com/departments/product-engineering/product/tools/feature_compatibility), the tables all overflow into the right-side menu. Curious if this might be related to https://github.com/sourcegraph/handbook/issues/789.

image (1)

jasonyavorsky:
Some page content really benefits from not having the sidebar. In that case, this option allows for hiding the sidebar (removing the table of contents and contributors) and moving the edit links to the bottom of the page.

### DIFF
--- a/src/pages/[...slug].tsx
+++ b/src/pages/[...slug].tsx
@@ -101,13 +101,16 @@ export default function Page({ page }: PageProps): JSX.Element {
         <>
             <NextSeo title={page.frontMatter?.title || page.title} description={page.frontMatter?.description} />
             <div className="container">
-                <nav id="right-sidebar">
-                    <section className="right-sidebar-section" ref={tocReference}>
-                        <h4 className="sidebar-heading">On this page</h4>
-                        <TableOfContents toc={page.toc} className="table-of-contents" />
-                    </section>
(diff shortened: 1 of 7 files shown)


src/pages/[...slug].tsx
```diff
@@ -101,13 +101,16 @@ export default function Page({ page }: PageProps): JSX.Element {
...
             <NextSeo title={page.frontMatter?.title || page.title} description={page.frontMatter?.description} />
             <div className="container">
-                <nav id="right-sidebar">
-                    <section className="right-sidebar-section" ref={tocReference}>
-                        <h4 className="sidebar-heading">On this page</h4>
-                        <TableOfContents toc={page.toc} className="table-of-contents" />
-                    </section>
-                    <EditSection page={page} />
-                </nav>
+                {!page.frontMatter?.hide_sidebar ? (
```
Comment (felixfbecker) [inline src/pages/[...slug].tsx:104]: tip: Instead of a ternary with `null`, you can just use `&&` (which is the conventional way to do it in JSX)

PR #2401, opened 2022-02-17
Scope strategy pages to next year or so

jasonyavorsky:
This updates the header on the Code Graph strategy pages to make it clear that they are scoped to within the next year or so.


jasonyavorsky:
Cc @yinquanteo

PR #2462, opened 2022-02-22
Mention that live coding exercise is done with React

lguychard:
@umpox @felixfbecker not sure who best to tag here, please redirect if appropriate -- we were reading over this with @limitedmage and realised we might want to clarify.


limitedmage (approved):

umpox (approved):
Thanks!

PR #2482, opened 2022-02-23
Remove key accounts deck link

malomarrec:
From what I understood, we are not using this deck anymore. If this is true, removing from product reference makes sense to not confuse teammates


christinaforney (approved):
Thanks Malo! I wonder if we should have an old/archived? So that folks could find this for historical reasons? We should also make it more clear that it was from FY22 Q4 if so? What do you think?

malomarrec:
@christinaforney proposed something. Wouldn't the diff be enough, though? I'm torned on this one

christinaforney (approved):

PR #2506, opened 2022-02-24
Replace "hiqh quality" by "high quality through iteration"

malomarrec:
**TL;DR - where does being iterative fit within our values? I miss having that as a principle.**

I really love our values, and they help me make decisions and build good habits day to day. For example, when we moved to customer first, it was great because it instantly aligned everybody on that principle and helped resolve discussions much more quickly. That change of value was 10x more impactful than a change of plan because it was instantly clear what it meant.

As we grow, we are planning more, and getting more structured. This is great! We have the chance to be at scale now and that means we can do ambitious things to help our customers.

One thing I've started to notice though is, given our scale and the more intense coordination required, we are tempted to produce bigger chunks at a time. Some of this is the way we build things (build and release big pieces), some of this is the way we market (hide features for a long time), some of this is the way we plan. Paired with a long release and upgrade cycle, it means that we sometimes don't get things to customers as fast as we used to. It also means that we are increasing our feedback cycle time and slowly getting farther away from customers.

In practice, this results in people at Sourcegraph more and more often asking themselves "is this sufficiently good to share/ship"? For everything communication-related, this is solved by being eagerly transparent, which is included in our Open and Transparent value. But for anything output/product-related, nothing in our values encourages us to build iteratively.

In the High Quality [section](https://handbook.sourcegraph.com/company-info-and-process/values/#high-quality), we briefly touch upon why high quality does not prevent us from being iterative. But nothing encourages us to be iterative either (emphasis mine).
> High-quality isn’t the opposite of iterative. You can keep quality high while still being iterative by narrowing the scope (for example, by solving a smaller problem at first, or by solving it only for certain customers initially).

So bottom line, sometimes I feel like we miss something that guides us about how big should a unit of value be before we build/release it and how much we spend on planning vs iterating. I don't know if this means we miss a value, or if this should be included in another value (eg. high agency).
Be iterative sounds like something we want to (a) avoid being too planning focused (b) avoid deadlocks (c) make sure what we do is de-risked ASAP.

**A technical example:**
TL;DR: Cut scope, ship fast, for a small number of users.

We want to improve Sourcegraph by allowing users to automatically create a Batch Change from an Insight. This is an enormous scope! We could also first create a button that just opens Batch Changes from Insights as a first step. But is that good enough? Most users will just open and do nothing, but some will create a batch change.

Being iterative means that we should make the smaller change first. That’s because it:
- doesn’t degrade the experience of most users
- can improve the experience of 1% of users a little
- allows us to get feedback

That doesn’t mean we leave things in a half-baked state, though. That would be a failure mode of being iterative. We need to have a strong vision to make that feature great in the future, and follow up on it. But the first step can and should be very small.

**A non-technical example:** 
TL;DR: Take small decisions instead of bundling lots of decisions in one conversation.

A good example of being iterative is to [avoid bundling](https://news.ycombinator.com/item?id=28463789) decisions together. It’s tempting to group a lot of projects together and try and solve them all the same time, because they’re related. But that means: it increases the perceived risk (massive scope), so it increases the number of people that want to look at it, which in turn increases scope (if we decide this, then we need to take that into account), which in turn increases risk. This means decisions risk being stuck or slow. Instead, where possible, don’t bundle decisions together.

This is strongly inspired by GitLab's [Iteration](https://about.gitlab.com/handbook/values/#iteration) value

#### Post-merge TODO

- [ ] :value-high-quality:  could be modified to represent a cycle


jeanduplessis:
@malomarrec I am very much in favor of adding this as one of our guiding principles. As we grow larger I believe being iterative is the antidote to slowing down and allowing us to stay nimble.

My only reservation in the current form of the diff is that I don't believe we should be adding more top-level values. We probably need to reduce them if anything.

As such, I'd highly recommend we add it as a sub-value of one of the existing ones. I think high-agency is the right value to add it under as it also reduces the risk of acting with high-agency if you do it in an iterative manner.

content/company-info-and-process/values/index.md
```diff
@@ -46,6 +46,24 @@ With agency comes responsibility; teammates are expected to take initiative and
...
+Having a strong vision and strategy allows us to be iterative: it gives us the framework that we need to move in the right direction, and allows us to plan less and iterate faster instead.
+
+In practice this means we:
+
+- cut scope to get things out faster
+- don't hide work in progress from customers as long as the change is better than the previous state
+- start a proposal, instead of waiting to organize a meeting to collect input
+- try to make reversible decisions, that allow us to change our mind if we need to
+
+Being iterative helps us have high agency, because small steps are easier to take and don't require approval. This is why it is a sub-principle of high agency. It also allows us to be more open and transparent, because we can learn in public by consyantly shipping small improvements.
```
jeanduplessis:
```suggestion
Being iterative helps us have high agency, because small steps are easier to take and don't require approval. This is why it is a sub-principle of high agency. It also allows us to be more open and transparent, because we can learn in public by constantly shipping small improvements.
```

jeanduplessis (approved):
Fully support this update to bring iteration into focus.

content/company-info-and-process/values/index.md
```diff
@@ -46,6 +46,24 @@ With agency comes responsibility; teammates are expected to take initiative and
 
 While [decisions should be made eagerly](../communication/decisions.md), this doesn't mean that you should make unilateral decisions when others should be involved. As an [open company](../about-sourcegraph/index.md#open-company), we value open discussion on important topics, and think that collaboration should be a large part of making things better. It is not a bad outcome if your proposal is not accepted or replaced with a better solution, because you took the initiative, started the conversation, and either reaffirmed that our current solution is good, or implemented an even better one! We're a startup, and if we take the safe path, we will most likely fail.
 
+#### Be iterative
+
+We try to ship small things fast, instead of big things slowly. Being iterative allows us to avoid unecesary long planning, avoid deadlocks, and derisk what we do faster, increasing the value we create for customers. Postponing shipping things leads to them being outdated, creates complexity in managing what's not been shipped, and delays learning.
```
rebeccadee:
```suggestion
We try to ship small things fast, instead of big things slowly. Being iterative allows us to avoid unnecessarily long planning, avoid deadlocks, and derisk what we do faster, increasing the value we create for customers. Postponing shipping things leads to them being outdated, creates complexity in managing what's not been shipped, and delays learning.
```

yinquanteo:
Agreed with the general idea of breaking work down to reduce time to value.

Where I see "be iterative" being problematic is for initiatives impacting large parts of any organization. Take planning for example – imagine we're iterating rapidly with our planning process and expecting folks to keep up with small incremental changes every day. In a scenario like that, it's with almost certainty that change fatigue will ensue. Communicating and keeping up with change is work in and of itself. And being iterative without the right change management processes in place can end up overwhelming people and have the opposite effect of slowing us down.

unknwon:
Personally, I favor very much "iterative shipping", as for planning, kinda agree that sometimes we do want to plan ahead enough until we can start building, depending on whether the decision is easily reversible or not.

unknwon (approved):
Thanks for doing this! ❤️

malomarrec:
@yinquanteo that's a good point. I guess even for planning, Open and Transparent would point towards eagerly sharing by default, but providing clear guidelines that nobody has to contribute or look at it until a set time, and avoid over-communicating in short cycles (so not being too iterative)? Is that what you mean?

Do you think this proposal should just be dropped, or can you suggest a short amendment to make it clearer?

yinquanteo:
@malomarrec: 
"Do you think this proposal should just be dropped". No! There's value in moving forward if we included some caveats.

Taking the planning process example we discussed, 
- Iterating successfully to me could look like: Having a small planning working group iterating on the planning process and batching up changes before informing/announcing to the broader population.
- Iterating unsuccessfully could look like: Big planning planning working group iterating on the process and expecting the broader population to iterate with them. 

In practice, we keep project groups small and provide clear guidelines on who should be involved when (as you mentioned). 

Have you seen these slides? https://komoroske.com/slime-mold/. I love the ideas around coordination and organizational headwinds.

jasonyavorsky:
I like @jeanduplessis 's idea of having iteration be a sub-value under high agency. Iteration helps you get things done because small changes avoid getting blocked in committee.

To @yinquanteo 's point, iteration worked pretty well at GitLab because the changes really were kept small, and people knew to make changes that were resilient such that the whole company didn't have to instantly adopt it in order for it to be successful. When I think about how we have iterated on planning, it has felt like throwing away much of the existing process and rewriting a big chunk, and then publishing that as a new iteration, and then doing that again. That is of course unsustainable and not really what I'd describe as iteration (some other things people think of as iteration but aren't really are listed here: https://about.gitlab.com/blog/2021/12/01/dont-confuse-these-twelve-shortcuts-with-iteration/).

I think the other thing that we had in our favor at GitLab was a handbook-first focus, so before anyone executed a process that was documented, you'd go read the latest version of the page that docuemnts it. No big change management or education process was needed, because the handbook was always the current documentation and you didn't need to keep up until you actually needed to do something. I don't think we exactly have that habit today.. but that said, you don't ever acquire that habit if you don't ever set the expectation.

jasonyavorsky:
Another thought is I don't think that working iteratively (when we are talking about product features, rather than planning) is compatible with embargoes so we'd need to work with marketing and product leadership (@christinaforney @amieroth and @abschu) to make sure we figure out how that process is going to work going forward if we want to make that more of a focus around how we do product development. Specifically - how would iterating on upcoming important features fast and publicly with customers be compatible with ensuring there is no public visibility on our upcoming important features?

We could connect this piece with the "how do embargoes connect with our 'transparency even when it is hard' value" discussion which I know is also ongoing.

content/company-info-and-process/values/index.md
```diff
@@ -46,6 +46,24 @@ With agency comes responsibility; teammates are expected to take initiative and
...
+
+- cut scope to get things out faster
+- don't hide work in progress from customers as long as the change is better than the previous state
+- start a proposal, instead of waiting to organize a meeting to collect input
+- try to make reversible decisions, that allow us to change our mind if we need to
+
+Being iterative helps us have high agency, because small steps are easier to take and don't require approval. This is why it is a sub-principle of high agency. It also allows us to be more open and transparent, because we can learn in public by constantly shipping small improvements.
+
+Being iterative allows us to work efficiently in an async-first setting. As long as we are aligned on our vision and strategy, being iterative reduces the need for heavy coordination, because we are mostly taking small reversible steps. It reduces communication bandwidth, because in many case we can share concrete artifacts (a PR, customer feedback on an iteration, a proposal) instead of sharing larger plans.
+
```
jasonyavorsky, line 67:
```suggestion

Because we're a handbook first company, the next time someone comes along and needs to follow the process that we've been iterating on, they can just read the handbook page and get caught up. We should avoid iterating in a way where it requires change management or large company-wide communication (for example, changing planning in the middle of planning) but apart from that, by being handbook first, we can also enable frequent and small iterative improvements.
```

jasonyavorsky: We should link to https://handbook.sourcegraph.com/handbook/#what-does-it-mean-to-be-handbook-first as well.

malomarrec: Thanks for the suggestion

malomarrec:
@yinquanteo I thought about your point about planning and discussed this with @jyavorska. Do you think your concern is addressed by the last version?

>We should avoid iterating in a way where it requires change management or large company-wide communication (for example, changing planning processes in the middle of planning).

> Because we're a [handbook first](../../handbook/index.md#what-does-it-mean-to-be-handbook-first) company, anyone can rapidly get caught up on recent changes by reading the handbook page.

content/company-info-and-process/values/index.md
```diff
@@ -46,6 +46,28 @@ With agency comes responsibility; teammates are expected to take initiative and
...
+We should avoid iterating in a way where it requires change management or large company-wide communication (for example, changing planning processes in the middle of planning).
```
jasonyavorsky:
```suggestion
We should avoid iterating in a way where it requires change management or large company-wide communication (for example, changing planning processes in the middle of planning). You can do this by either waiting until the end of the process, or by making it clear that the changes will take effect next time.
```

virginiaulrich:
@malomarrec interestingly (well, at least to me), I could also see this fitting with continuously grow. I leave it you decide how to best reflect what you are seeking to accomplish with this, of course. My thought is that iteration is inherently what drives anything that is continuous (maybe I have too many hears doing LEAN in factories to not think this way). I also think that this living as part of continuously grow allows for a more healthy tension between values. High agency without teamwork is just bulldozing others. Or, in this case, it could be ... high agency without continuously grow is just doing something quickly and never thinking about it again.  Again, please do proceed as you see fit -- these are just the thoughts that came to mind.

malomarrec:
Thanks for the thought @virginiaulrich, and I agree. This subvalue fits under multiple top-level values, and continuously grow is a good one (I think @sqs said something like "this is also about 'learn in the open'").

content/company-info-and-process/values/index.md
```diff
@@ -46,6 +46,28 @@ With agency comes responsibility; teammates are expected to take initiative and
...
 
+#### Be iterative
+
+We try to ship small things fast, instead of big things slowly. Being iterative allows us to avoid unnecessarily long planning, avoid deadlocks, and derisk what we do faster, increasing the value we create for customers. Postponing shipping things leads to them being outdated, creates complexity in managing what's not been shipped, and delays learning.
+Discussions around large changes tend to be slow and deadlock, whereas small steps are much more likely to lead to a decision. It's always easier to re-evaluate after we have learnt by taking a first step.
```
jasonyavorsky:
```suggestion

Discussions around large changes tend to be slow and deadlock, whereas small steps are much more likely to lead to a decision. Bundling lots of problems together in the hope that you can fix them all at once is attractive, but it rarely works and as you increase the complexity of your proposal it becomes more and more likely that the whole thing will get stuck. It's also always easier to re-evaluate after we have learnt by taking a first step.
```

malomarrec:
Another thought that came up in a conversation with @mrnugget (but I am not representing his opinion here, just my own, so  credits to Thorsten if he agrees with it, all responsibility on me for being a poor scribe if not): 

It seems to me that **high quality alone is not very useful as a value because it is quite low signal-to-noise**. Shipping high quality is not controversial, nor is defining what good quality is. Adding be iterative there makes it much more useful.

By contrast, other values are high signal-to-noise in the sense that they help to make sense out of situations where the course of action is unclear:
- Be iterative is pretty clear: you just ship small things until there is a clear business reason to do so. This is high signal-to-noise: it's not natural to be that way.
- Open and transparent: you make things public until there is a clear business reason to do so. This is controversial: it's not natural to be that way. High signal again.

Values should be helpful to make sense of difficult situations, otherwise they risk becoming this vague abstract overarching concept. When reading the high quality page I don't see much that helps in practice or would settle a controversial conversation.

jasonyavorsky:
That's an interesting point @malomarrec but I wouldn't add it to the PR or it might become too complex. I would treat improving the high quality value as a separate transaction, which de facto replaces it with this one (but avoids this getting stuck in the mud). Since it's not harmful for both values to exist for some time, it can help keep each PR simple.

mrnugget:
I'm wondering whether this newly proposed value isn't better version and replacement of "High Quality".

The first sentence describing the "High Quality" value:

> High Quality
> You are responsible for finding out what high-quality work looks like and **delivering that high-quality work iteratively**.

Emphasis mine. 

As Malo wrote in the previous comment: I do think "High Quality" is a low-signal **name** for this value.

Because what we mean with "High Quality" ("finding out what high-quality work looks like", meaning it's subjective) is counter to common ideas of "high quality" (if we'd ask a 100 people what "high quality" means, I bet a lot would say "something's perfect", "really good", etc.). We use "High Quality" to also include ideas like minimal-viable products, prototypes, experiments, making trade-offs, etc. What exactly is of high quality depends on the context and what we're trying to do - that's what this value says to me, in its *description*, not in the name.

So, yes, I'm being completely unhelpful here by disagreeing with the previous ideas of grouping it under "high agency" or "continuously grow" and adding another opinion: I think we should rethink the "High Quality" value and possibly merge it with the description here and rename it. Sorry! 😄

jasonyavorsky (approved):

malomarrec:
GitLab specifically calls out iteration as a reason for success in their latest [earning call](https://seekingalpha.com/article/4495480-gitlab-inc-gtlb-ceo-sid-sijbrandij-on-q4-2022-results-earnings-call-transcript)

> Another reason for our fast innovation is our value of iteration. Iteration means reducing the scope of the next piece of work with the smallest valuable thing possible, so it can be delivered fast. Our innovation is also predictable as we shipped on the 22nd day of every month for 124 months in a row as of January. This allows our customers to benefit from new innovation every month and it allows us to get feedback very rapidly.

>To provide some examples of this innovation, in the fourth quarter, we delivered enhancements in the following areas. First, software supply chain security continues to be a key focus for enterprises and providing controls that empower teams is critical as this helps everyone be more secure. With this in mind, we improved fine-grained security approval controls and merge request approval controls by enabling every software change to be reviewed by at least two people. This allows organizations to better manage their security risk. Second, digital transformations continue to drive cloud-native adoption and we see this as a key enterprise initiative for quite some time. To support this, we are continuing to expand our cloud-native capabilities, including improving the user experience for a GitLab agent for Kubernetes and infrastructure as code security scanning capabilities. Third, we recognize that GitLab is a critical business application for our customers, whether they are new start-up with a small team or enterprise customers with tens of thousands of users. Thus, we continue to invest in improving both the performance and scalability of our DevOps platform. We recently made improvements to our GitLab container registry to enable better performance while also reducing storage costs. Additionally, we made sure that our geo capabilities work seamlessly worldwide with better performance. Fourth, we improved our core functionality, including source code management, CI/CD and security features. These investments support our typical customer journey. As to date, our customers tend to start with source code in CI, then expand to use secure package and release.

nicksnyder:
Just want to share some historical context on this. When Christina and I originally drafted these values, our thinking was `high quality through iteration`. 

You can see the definition of high quality already mentions iteration:

> [High quality](https://handbook.sourcegraph.com/company-info-and-process/values/#high-quality): You are responsible for finding out what high-quality work looks like and producing that high-quality work iteratively.

At one point we did have an intention of changing `high quality` to `be iterative` and you can see the PR here:
https://github.com/sourcegraph/about/pull/1738

That part of the PR was reverted because we couldn't find wording that we were happy with and we wanted to push through the other changes.

[The follow-up PR to actually reconsider changing](https://github.com/sourcegraph/about/pull/1760) `high quality` to `be iterative` never left the draft state for reasons I can't remember (I assume we got busy with other things and forgot about it).

In any case, I agree that:
1. We should avoid net-adding new values if possible
2. A value about being iterative is more valuable than one about quality (as evidenced by those previous PRs) so I would be in favor of doing that replacement.

malomarrec:
@nicksnyder - thanks for the added context. I took your suggestion and proposed merging high quality with being iterative, reusing the "High quality through iteration" wording.
The benefit of that is that the key problem @mrnugget identified, which is that the hard part is _figuring out what quality is_, is now better defined: _you can do that through iteration_.

That also meant that some of the recommendations in High quality could be removed or simplified. For example, "If you don't have the time to build high quality, cut the scope" is natively implied by being iterative. The one thing I am not sure about is what to do with:

> If you don’t have the skills to produce high-quality work, then it's your responsibility to acquire those skills (for example: learn them yourself, make use of [Sourcegraph's education budget](../../benefits-pay-perks/benefits-perks/travel/index.md#5-professional-development-stipend-and-educational-travel), delegate to an existing teammate with the necessary skills, or hire a new teammate with those skills).

It feels like is a high agency type advise and can be safely removed, but I'm not certain.

jasonyavorsky:
I really like where this has ended up @malomarrec @nicksnyder

nicksnyder (approved):

sqs:
I like this change a lot (high quality through iteration). I think this captures what we ideally want better than what exists today in the high quality value. It would also have been useful for me to refer people to several times in the last few weeks alone. Thank you for your work iterating on our values to make them higher quality. :)

I want to share this proposed change (this PR) with the company at next company meeting. It seems like a change to values should be mentioned there before being merged, to make sure as many people as possible are aware of it and can give feedback. Along with that we will post to #announcements. Then a few days later we can merge (unless serious issues are raised). @madisongclark Can you please add this to the next company meeting (working with @malomarrec)?

This week I will make a PR to the handbook documenting how I think values changes should go (basically open a PR, get feedback from anyone interested, then w/my approval share it in company meeting and #announcements and wait a few more days for feedback, then merge--which is exactly what is going on here).

madisongclark:
Noted, we'll add this to the agenda for the next company meeting. I'll connect with you directly @malomarrec so we can align on the talking points.

chrispine (approved):

nicksnyder:
I am updating the summary line here: https://github.com/sourcegraph/handbook/pull/3045

PR #2508, opened 2022-02-24
Add a few extra checks during the managed instance upgrade process

caugustus-sourcegraph:
Quick attempt to improve the documented process while the larger discussions are ongoing.


kevinwojo (approved):
This is good! Thanks!

michaellzc (approved):
good stuff

PR #2559, opened 2022-02-28
Add analytics FAQs

ebrodymoore:


content/departments/bizops/analytics/faqs.md
```diff
@@ -0,0 +1,33 @@
+# Data & Analytics FAQs
+
+Below are answers to common questions teammates at Sourcegraph have.
+
+### Current seat limit, ARR, company type, and more are outdated on the [Server Instances Overview](https://sourcegraph.looker.com/dashboards/167). How can we get this updated?
```
kelsey-brown: Joe pulled the current_seat_limit field from salesforce so it's actually just not an option anymore. May want to remove it from this line to avoid confusion that we can currently access that information in Looker. I'll add to the source of truth doc the places you can find license seat count.

content/departments/bizops/analytics/faqs.md
```diff
@@ -0,0 +1,33 @@
...
+
+Looker only populates this information from Salesforce for customers, and Looker and Salesforce both are not connected to our licensing system, so trials as a `Company Type` can't be picked up. The Data & Analytics team can't update any of this information - instead, please update the relevant information in Salesforce and it'll be updated in Looker within 24 hours.
+
+### My account's [insert metric] went from really high to really low week over week. What happened?
+
+It's likely they have multiple instances (such as a separate instances for staging and production) and our data pipelines picked up the low data point from their staging instance. Unfortunately fixing this is very difficult and not among our top priorities, so please reach out in #analytics to confirm specific pieces of information.
+
+### I want to know which users did this action in Amplitude. Can I pull their email addresses?
+
```
kelsey-brown, line 14: I searched docs and didn't find anything, so I think the answer to this question is no but asking just in case: does doing this type of thing contradict what we say we do with users' PII in any of our docs or agreements? Just flagging because if so, we should probably update the docs

kelsey-brown (approved):
A couple comments but looks good

content/departments/bizops/analytics/faqs.md
ebrodymoore: No we're good here!

PR #2584, opened 2022-03-02
Separate `problem` and `solution` review steps in RFCs

jasonyavorsky:
A pattern I've noticed is that some RFCs jump into detailed solutions prior to the point that the problem that's being solved is well understood; sometimes the problem doesn't actually exist, sometimes the problem is just a little bit different from what was originally there.

This PR separates solutioning from validating the problem so that we can be more efficient and avoid getting ahead of ourselves too much: in situations where the solution is written too early, there's often a ton of debate on the details of the solution that would have been more managable if it was around the problem.

This came from a discussion with @lguychard and I think it's a great idea.

When merging this, we can also update the RFC template itself to make it clear that it is often a good idea to wait until there is agreement on the problem before designing the solution.


content/company-info-and-process/communication/rfcs/index.md
```diff
@@ -64,7 +64,8 @@ These status labels help to provide more context for readers. They’re most hel
 We have a set of status labels we’ve found helpful in the past:
 
 - **WIP**: The author is still drafting the RFC and it’s not ready for review.
-- **Review**: The Review label is used when the RFC is ready for comments and feedback.
+- **Reviewing problem**: The Reviewing problem label is used when the RFC has a firm problem statement and is ready to confirm that the problem definition is correct.
```
lguychard:
```suggestion
- **Reviewing problem**: The Reviewing problem label is used when the RFC has a firm problem statement, and reviews should focus on confirming that the problem definition is correct.
```

lguychard (approved):

malomarrec (approved):
Great addition. I find myself often doing that, and it'd be great to have a convention around it.
Potentially clarify that having a good problem statement is orthogonal to appetite and timing, so that it's clear it's not a matter of validating priorities but of validating problems.

jeanduplessis (approved):

yinquanteo (approved):

jasonyavorsky:
Merged, I also added some text to the template with the same advice.

PR #2616, opened 2022-03-04
Fix generated absolute links

jasonyavorsky:
The trailing slash detection fix (https://github.com/sourcegraph/handbook/commit/d4aa8316f5dc3b5febd8cec65686b3c0044ddaa8) was resulting in all generated links having an extra `../` in their URLs, making them invalid. This corrects it.


jasonyavorsky:
@katjuell I am not sure if this re-breaks the trailing slash fix - hopefully not, but can you please take a look?

katjuell (approved):
LGTM! I tested some of the problematic use cases we knew of from the trailing slash reported issues, and everything looks good! I think we can close #2620 and #2607 after merging this.

jasonyavorsky:
Great, thank you @katjuell !

PR #2627, opened 2022-03-07
Add note about re-referencing handbook

jasonyavorsky:
At GitLab, which is also a handbook-first company, one of the ways we reduced lots of Slack and other communication around what is changing in the handbook all the time is that everyone has a habit of checking the handbook before you do something that's documented there, just to make sure you're up to speed on what's new. This of course also relies on the handbook being the source of truth, and up to date, but that's another topic.

For example, if you were creating an expense report, you wouldn't just rely on your memory of when you read it in the past, but do a quick check on the handbook page to see if it's changed. Probably not every time, but at least once every few expenses you submit.

I think if we had more of a habit of referring to the handbook regularly, we would have to worry less about manually communicating changes to the handbook in Slack or other places, and this would help us be more async and reduce communication noise.


jasonyavorsky:
I think this is non-controversial so merging.

PR #2652, opened 2022-03-08
Move local SVGs to public folder and update MD tags

### LINKED ISSUE #2657: Fix issue with SVGs not loading properly in Markdown

katjuell:
Next.js [recommends](https://nextjs.org/docs/basic-features/static-file-serving) serving static content from the `public` directory, so let's move our local SVGs there.

This PR should also:
- Update `object` elements in MD files to be `img` elements.
- Update documentation relating to SVGs.

It will close #2657 

## To Test
- Check out the following pages, and ensure that everything looks good with SVGs loading: `company-info-and-process/communication/asynchronous-communication`, `departments/marketing/brand/brand_guidelines/brand-expression`, `departments/marketing/brand/brand_guidelines/color`, `departments/marketing/brand/brand_guidelines`, `departments/marketing/brand/brand_guidelines/logo`, `departments/people-ops/process/impact-reviews`, `departments/product-engineering/engineering/code-graph/code-insights/processes`.
- Ensure that the new directions make sense at `handbook/editing/handbook-images-video`
- Here are some of the queries I used to test for [`object` elements](https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/sourcegraph/handbook%24+object+lang:Markdown+&patternType=literal) and stray references to the [`logo` SVG directory](https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/sourcegraph/handbook%24+logo&patternType=literal) — let me know if I've missed something!


public/static/project_planning.svg
```diff
@@ -1,4 +1,4 @@
-<svg viewBox="0 0 1100 263" width="1100" height="263" preserveAspectRatio="xMidYMid meet" xmlns="http://www.w3.org/2000/svg">
+<svg viewBox="0 0 1800 263" width="1100" height="263" preserveAspectRatio="xMidYMid meet" xmlns="http://www.w3.org/2000/svg">
```
felixfbecker, line 1: What's the thought behind this change?

content/company-info-and-process/communication/asynchronous-communication.md
```diff
@@ -129,7 +129,7 @@ Finally, remember Slack is _not_ a source of truth. Something said or mentioned
 We've talked above about how to do things async, but sometimes you might need (or think you need) a sync meeting. This section can help.
 
 <center>
-<object role="document" width="1133" height="1889" title="One pager: Async versus sync communication" data="onepager-async.svg" style="width: 80%"></object>
+<img width="1133" height="1889" title="One pager: Async versus sync communication" src="/static/onepager-async.svg" style="width: 80%"></img>
```
bretthayes, line 132: 1. Can we self close the `` elements?
2. Can we also update the img elements with alt attributes? This may be out of scope for this PR but it would be good for SEO.

felixfbecker (changes requested):
Thanks for looking into this! 😊 

> Next.js [recommends](https://nextjs.org/docs/basic-features/static-file-serving) serving static content from the public directory, so let's move our local SVGs there.

We run a script before every build that copies all assets over to `public` automatically, so that we can colocate pages and their assets like SVGs:

https://github.com/sourcegraph/handbook/blob/main/src%2Fscripts%2Fcopy-assets.ts

I think colocation is better for finding files, structuring folders and understanding the URLs to link them. It also avoids related images not being deleted when e.g. a directory of pages is deleted.

Is something broken with the script? Or is this related to the trailing slash behavior (that would be my guess)?

> Update object elements in MD files to be img elements.

Is there a motivation for this change? The content of an SVG in an `img` element is not exposed to accessibility tools (screen readers), so switching from `object` to `img` makes the SVG diagram  inaccessible.

felixfbecker:
Could this be related to CSP? I'm seeing these errors in the console:

```
[Error] Refused to load [...].svg because it does not appear in the frame-ancestors directive of the Content Security Policy.
```

Did we change our CSP? It seems like the proper fix would be to configure our CSP to include ourselves in `frame-ancestors`

public/static/project_planning.svg
katyhoward: I thought that this looked funky when I was as able to load it as is. Here's a screenshot from the #2660 deploy environment:
Screen Shot 2022-03-09 at 9 09 13 AM

katyhoward: Change looks like this:
Screen Shot 2022-03-09 at 9 21 36 AM

katyhoward:
@felixfbecker I'm having trouble loading all of the paths with local SVGs on the CSP deploy right now (have only been able to load `departments/product-engineering/engineering/code-graph/code-insights/processes` on #2660 for a brief moment), but it seems like that PR will resolve the issue? Thanks for calling my attention to that build script! I had missed that. 

Happy to close this out and open a separate PR with the SVG `viewBox` change above, keeping the `object` elements and current local asset locations in place.

bretthayes (changes requested):

public/static/project_planning.svg
felixfbecker: The cut-off is intentional, each project is the same steps, so it's to symbolize that it repeats endlessly to the right. The sizing is so the font size is legible (matches the font size of the text body), if it's resized the text would become too small.

PR #2660, opened 2022-03-09
Extend CSP

felixfbecker:
This allows resources from the handbook to be embedded in the handbook itself. Specifically, we need this for `<object>` embeds of SVGs.

It also moves a `<script>` to an external file, because we have `'self'` but not `'unsafe-inline'`.


katjuell (changes requested):
Awesome! Unfortunately, Netlify is being [flaky](https://twitter.com/netlifystatus/status/1501551460936413187?s=20&t=s3qnZw4FdjALPro4vx2aPw) this fine morning, so I've only been able to load `departments/product-engineering/engineering/code-graph/code-insights/processes` so far, but it looks like the logo is loading. I just see a couple of warnings in the console
Screen Shot 2022-03-09 at 9 08 35 AM

felixfbecker:
Those seem like separate issues (domains that should be additionally added to the CSP `style-src` directive). I can fix these too in this PR.

katjuell (approved):
lgtm!

PR #2678, opened 2022-03-10
Reduce meetings for new teammates

rrhyne:
As written the new teammate would need to meet with a prohibitive number of people.

Replaces usertesting.com purposefully.


marisak (approved):
Looks great! I was definitely overwhelmed by the number of people on the "to-meet" list so this should help a lot.

Joelkw (approved):
Thank you @rrhyne !

PR #2680, opened 2022-03-11
process: update external contributions guide

bobheadxi:
Provide better instructions for requesting a build, and add link to what is probably a duplicate document, but oh well

Saw this page mentioned in https://github.com/sourcegraph/sourcegraph/pull/32360#issuecomment-1063710719


unknwon (approved):
I knew we had some `sg ci` magic!

PR #2722, opened 2022-03-15
Add embargo and sanction countries

mercadon:


mercadon:
@jonathanayers Any concerns on merging this into our handbook?

tammy-zhu (commented):
@mercadon we can remove Sudan, since the US took Sudan off the embargo list.

PR #2733, opened 2022-03-16
Add definition of technical acronyms

IAmHughes:
I noticed throughout the glossary, there were a few terms that had acronyms in their definition and/or were explanations of the service/product, but the acronym itself was never explained. This expands on some of the definitions for additional clarity.


IAmHughes:
Looks good to me! :shipit: 

image

InesRoitman:
This is amazing Thomas! Thank you!

PR #2742, opened 2022-03-17
Slack Hanbook Update- adding new channels

toddherskovitz:
adding a few new channels to the Slack section of the handbook


toddherskovitz (commented):
updated commit message name

PR #2745, opened 2022-03-17
Update locations.geojson > Add Becca Steinbrecher

st0nebreaker:
Spacing is off because the last addition to this json was out of order ([PR#2622](https://github.com/sourcegraph/handbook/pull/2622)) & broke the json. Used jsonformatter.io to fix the small bug & add my own entry.

Let me know if the spacing/indents are important here to leave as they were!


bretthayes (commented):
✅ Nice work! Thanks for catching this! Now the map renders again! 🎉

PR #2762, opened 2022-03-21
Add CloudSQL backup/restore to incident playbooks

filiphaftek:
The purpose is to have all documentation in one place during the incident.

### DIFF
--- a/content/departments/product-engineering/engineering/process/incidents/playbooks/index.md
+++ b/content/departments/product-engineering/engineering/process/incidents/playbooks/index.md
@@ -189,7 +189,7 @@ You can also perform these commands from the [Google Cloud SQL UI](https://conso
 
 ### Restore database using point-in-time recovery (creates new database clone!)
 
-🚨 You should open an incident (unless it is already open) and notify the #dev-ops channel if a situation arises when a point-in-time recovery is required.
+🚨 You should open an incident (unless it is already opened) and notify the #dev-ops channel if a situation arises when a point-in-time recovery is required.
 
 Explanation of [CloudSQL PostgreSQL point in time recovery](https://cloud.google.com/sql/docs/postgres/backup-recovery/pitr)
 
--- a/content/departments/product-engineering/engineering/process/incidents/playbooks/index.md
+++ b/content/departments/product-engineering/engineering/process/incidents/playbooks/index.md
@@ -187,6 +187,21 @@ You can also perform these commands from the [Google Cloud SQL UI](https://conso
 
 🚨 You should notify the #dev-ops channel if an situation arises when a restore my be required. It should also be filed in our ops-incident log.
 
+### Restore database using point-in-time recovery (creates new database clone!)
+
+Explanation of [CloudSQL PostgreSQL point in time recovery](https://cloud.google.com/sql/docs/postgres/backup-recovery/pitr)
+
+Before starting restoring to point in time, ensure that previous step with restore via daily backup is not possible - point in time operation has to create new database instance and all applications have to be reconfigured to use new instance!
+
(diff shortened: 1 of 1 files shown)


sanderginn:
What's the intended course of action when the database is unable to be backed up (e.g. due to corruption) and the most recent backup is older than 1 hour?

filiphaftek:
> What's the intended course of action when the database is unable to be backed up (e.g. due to corruption) and the most recent backup is older than 1 hour?

If we cannot accept to restore to the last backup (b/c it's too old), then imho we would have to perform [point in time recovery](https://cloud.google.com/sql/docs/postgres/backup-recovery/pitr) - but this is way harder, as we need to switch applications to use new instance.

@daxmc99 @danieldides what is your suggestion here ^^ ?

daxmc99 (approved):

content/departments/product-engineering/engineering/process/incidents/playbooks/index.md
```diff
@@ -175,6 +175,33 @@ Open the alert UI to click on the check URL that was failing and verify it's now
...
+Before starting restoring to point in time, ensure that previous step with restore via daily backup is not possible - point in time operation has to create new database instance and all applications have to be reconfigured to use new instance!
+
+You can create a clone of the instance from point in time via `gcloud sql instances clone ${instance_name} ${clone_instance_name} --point-in-time '2022-03-21T19:30:00.000Z'` (UTC timezone for the source instance in RFC 3339 format).
+
+You have to create a PR which modifies all CloudSQL Proxy conigurations to point to new instance in [base folder of the repository](https://github.com/sourcegraph/deploy-sourcegraph-cloud/tree/release/base)
+i.e. `sourcegraph-dev:us-central1:${instance_name}` -> `sourcegraph-dev:us-central1:${clone_instance_name}`
+
+After PR is approved and merged, applications are redeployed and use new instance.
+
+🚨 You should notify the #dev-ops channel if an situation arises when a restore my be required. It should also be filed in our ops-incident log.
```
daxmc99: Suggestion
Move this to the top and suggest that an incident is opened. If a restore is needed, we should be aware.

content/departments/product-engineering/engineering/process/incidents/playbooks/index.md
```diff
@@ -175,6 +175,33 @@ Open the alert UI to click on the check URL that was failing and verify it's now
...
+
+You can also perform these commands from the [Google Cloud SQL UI](https://console.cloud.google.com/sql/instances?project=sourcegraph-dev)
+
+🚨 You should notify the #dev-ops channel if an situation arises when a restore my be required. It should also be filed in our ops-incident log.
+
+### Restore database using point-in-time recovery (creates new database clone!)
+
+Explanation of [CloudSQL PostgreSQL point in time recovery](https://cloud.google.com/sql/docs/postgres/backup-recovery/pitr)
+
+Before starting restoring to point in time, ensure that previous step with restore via daily backup is not possible - point in time operation has to create new database instance and all applications have to be reconfigured to use new instance!
```
daxmc99:
```suggestion
⚠️ **WARNING:** Before starting restoring to point in time, ensure that previous step with restore via daily backup is not possible - point in time operation has to create a new database instance and all applications have to be reconfigured to use new instance!
```

From https://sourcegraph.com/github.com/sourcegraph/handbook/-/blob/content/departments/support/process/removing-repositores-from-cloud.md?L9:6 
Renders like [this](https://handbook.sourcegraph.com/departments/support/process/removing-repositores-from-cloud/)

daxmc99:
Honestly, it depends on the circumstances. In the vast majority of cases, I expect we would choose the data loss over needing to reconfigure the application and the time a point-in-time recovery might need. 

Number 1 use-case here is to replicate an incident exactly. Database state before something occurred.

PR #2778, opened 2022-03-22
core: precise wording of monorepo performance goal

keegancsmith:
We now have a repository which fits the bill of our older performance goal: gigarepo. Lets directly reference it to make the goal easier to communicate. I include what the gigarepo is for people not familiar with the repository. Additionally, I found the wording "15GB repo" unclear since there are many ways to interpret that. So this specifically states it is the HEAD working copy size. IE it excludes the .git folder size.

Maybe we can explain what gigarepo is somewhere else rather than in the bullet points? For now this makes it more self contained. cc @sourcegraph/search-core 

Co-Authored-By: @stefanhengl


tsenart (approved):
LGTM

eugeniaft (approved):
Looks good! Maybe we could add a small intro section before the goals and add it there? I can do that next.

PR #2822, opened 2022-03-25
Add offline instances source of truth

kelsey-brown:
Adding the offline instances doc, like we talked about yesterday. Let me know if you think there's another good place to link the doc? This seemed like the place where people would be most likely to see it.


ebrodymoore (approved):
👍 nowhere else I can think of to put it

PR #2838, opened 2022-03-28
Add pre-launch mentorship program edits

courier-new:
I've tried to incorporate portions of the original announcement draft into the handbook page, as well as answer questions I was commonly getting from folks I shared the announcement draft with for review. I tried to clarify that compared to many traditional mentorship programs where there's a formal match-making system and a uniform duration/timeline for the program, our program works more async on a rolling basis, with mentees initiating the "match-making" themselves. I also updated the formatting of the "For mentees" section to make the steps a bit easier to scan.


courier-new:
(I'm going to merge this now just so it's live but feel free to still leave any comments/review and I can address it in a follow-up! 🙇‍♀️)

PR #2904, opened 2022-03-31
Remove documentation about old security infrastructure

ferozsalam:
We don't run the majority of infrastructure that is currently described.
Remove it pending a refresh.


deflncha (approved):
LGTM 👍

PR #2932, opened 2022-04-01
Update managed instance docs

andreeleuterio:
We're rolling out proxying through Cloudflare for managed instances and this comes with some doc updates.


daxmc99 (approved):

daxmc99:
Part of https://github.com/sourcegraph/infrastructure/pull/3022

PR #2937, opened 2022-04-01
update travel protocols with larger meet up

cassiemelani:


cassiemelani:
Team - updating this page on behalf of **[Coury's question](https://sourcegraph.slack.com/archives/C02FSM7DU/p1646747837591499)**

I haven't heard any direct questions/concerns from him on a specific trip, so I think he just wanted to see it updated. As a note, I've deleted the travel comfort survey that Debbie had created because we do not have access to see responses and have advised on how to create one for individual groups.

PR #2976, opened 2022-04-05
updated policies page with the computer policy

nickyvm:
I updated this page with the computer policy link


content/company-info-and-process/policies/index.md
```diff
@@ -1,18 +1,21 @@
 # Company wide directives and policies
 
-## Teammate privacy and trust
+## Directives
```
nicksnyder: what is the difference between a directive and a policy?

nickyvm: It isn't part of our SOC 2 policies so I felt weird adding it below. You and Dora suggested I add it here but doesn't seem like an official policy document. 

I see directives as the specifics that are shaped by the policy. Ie the computer requirements are shaped by the asset management policy but aren't explicitly stated there. I think eventually they will be merged as the 'policies' need to be re-written in Sourcegraph languange.

nicksnyder: Maybe I am missing something and "policy" is a term of art that means something special in compliance land, but I would casually call our device requirements a "policy" and would find it less confusion if they were just all grouped together as such.

nicksnyder: Maybe I am missing something and "policy" is a term of art that means something special in compliance land, but I would casually call our device requirements a "policy" and would find it less confusion if they were just all grouped together as such.

nicksnyder: I don't feel strongly about this. I defer to you on what you think makes sense here (either leaving it separate or not adding it to this list at all)

nicksnyder (approved):

nickyvm (commented):
I agree it was wordy and unnecessary. I've added it to the list

PR #2977, opened 2022-04-05
Add verbiage about feature owners

unclejustin:


unclejustin:
@CristinaBirkel this is my sad start to your feature owner idea 😂 

@coury-clark asking you to take a look at this because you often have fantastic input about these types of things. Please feel free to ignore if you do not have time right now.

CristinaBirkel:
@coury-clark had a question earlier about where we track feature owners. I proposed that maybe we use tracking issues. The feature owner would create the tracking issue and assign themselves, and then all of the other issues would be tracked there. What do you think? That would probably be a process to document here.

Here's kind of my take on what a feature owner might be responsible for:
1. Have a full understanding of the requirements and the work that needs to be done across the frontend and backend. The main benefit here is catching any misunderstandings or integration issues before they happen.
2. Keep an eye on progress and be proactive about raising blockers or delays.
3. Maybe the feature owner should also own the testing effort? This might look like manual testing on `k8s`.

Just some additional thoughts!

coury-clark:
@CristinaBirkel @unclejustin it kind of feels like we are asking the "feature owner" to be a full project manager, which I don't think is feasible or practical. Can someone clarify for me what the intention is here? Are we looking for a project manager, or are we looking for a single point of contact developer?

CristinaBirkel:
@coury-clark @unclejustin I see a significant difference between feature owner and PM. A couple of thoughts:

- If there's an issue with the functionality, (something isn't feasible or an edge case was missed, etc,) Joel will still be the decider on what to cut/adjust.
- If there's a significant delay, Joel will still decide what we want to get into the next release vs. push out.

The reason I suggested the 3 things above is:

1. We had a significant issue with LAM that could have been caught much earlier if someone had an understanding of the whole feature including the frontend and backend.
2. I think this is an area we can improve upon, especially around releases. It can get a little chaotic trying to keep track of whether a larger effort is on track when we have multiple people working on their own things. So my suggestion here was for somebody to stay on top of it as a whole, and I think the feature owner could make sense since that person would have a decent understanding of all of the tickets.
3. Using LAM as an example again, I kind of realized this week that we'd pushed all the changes to `k8s`.. but had anyone tested it? Whose responsibility is that? I think the feature owner would be a natural fit.

I don't see it as a PM. More of a transient dev lead kind of thing. It just seems like a gap that we have.

coury-clark:
@CristinaBirkel to be clear, I mean *project* manager, not product. These responsibilities take a fair amount of time and are generally non-trivial. I'm not saying it's not valuable, but I feel that putting this role on every dev on the team is going to result in a lot of overhead.

CristinaBirkel:
@coury-clark Ah you're right, I always get those confused. 😅 I agree that it's non trivial and time consuming. But I think that no one is really doing these things right now and that it's been causing issues. To me it seems like time well spent. And that it will possibly allow for more focus instead of less. Instead of everyone trying to understand everything, (and not having time for that, hence the gaps,) we can trust that the feature owner has got it covered.

content/departments/product-engineering/engineering/code-graph/code-insights/processes.md
```diff
@@ -55,6 +55,8 @@ Our current process is as follows:
 
   - We hold a [fortnightly synchronous iteration planning](#fortnightly-iteration-planning), we take a look at the proposed issues together on the ["Next iteration" view](https://github.com/orgs/sourcegraph/projects/200/views/4).
 
+- Larger issues that involve multiple domains may benefit from a "feature owner". A feature owner assumes full responsibility for gathering all requirements to complete an issue. The feature owner is expected to have full understanding of the issue and pull in any resources or team members needed to complete the task.
```
felixfbecker: @Joelkw's PR called this "project lead", should probably align the wording with that so it's consistent within the same handbook page (clear that it refers to the same person).

unclejustin:
@CristinaBirkel @coury-clark reviving this issue.

Quick update first: "feature owner" is now "project lead"

Coury, I think your concern about the time commitment would be alleviated by only assigning a project lead to larger issues.

IMO if we are assigning project leads to issues every iteration then we probably need to seriously look into hiring for a new dedicated PM role. The way I see this being implemented it should be relatively rare that someone really needs to lead a project like this. If we constantly have this need, then either we need a dedicated person or we are not breaking down our issues well enough.

chwarwick:
In general I think it's good to do with larger more complex items.  Is this specifically an engineering responsibility? If so is the assumption that when you take on this role the issue(s) are in a ready to be worked state or does it imply that getting it to the ready state is part of the responsibility.  

The first section makes it sounds like it's about getting it though the rfc/design stages, but the responsibilities section sounds more like ensuring the development work gets completed.

content/departments/product-engineering/engineering/code-graph/code-insights/processes.md
```diff
@@ -59,6 +59,24 @@ Our current process is as follows:
 
 - If one or more issues that were planned for an iteration are looking to **not get finished** (which includes testing) in the [current iteration](https://github.com/orgs/sourcegraph/projects/200/views/1) (while maintaining sustainable work practices) the assignee **raises this as soon as possible asynchronously** to the team (including the PM and EM), e.g. on the GitHub issue or Slack. These issues then **become _proposed_ issues for the next iteration** (meaning nothing carries over automatically, but we also don't just drop and forget missed issues).
 
+## Project Leads
+
+Larger issues that involve multiple domains may benefit from a "project lead". A project lead assumes full responsibility for gathering all requirements to complete an issue. The project lead is expected to have full understanding of the issue and pull in any resources or team members needed to complete the task.
```
Joelkw: Do we want to call these "issues" or "projects"? (Confusion with GH issues, or is this intentional/there will be per project issues as well?) (Update: I see we're using tracking issues, so this point is moot!) 

Also, do they necessarily have to involve multiple domains? (Monorepo improvements feels like primarily a backend project, for example?) 

```suggestion
Larger issues, especially ones that involve multiple domains, may benefit from a "project lead". A project lead assumes full responsibility for gathering all requirements to complete an issue. The project lead is expected to have full understanding of the issue and pull in any resources or team members needed to complete the task.
```

content/departments/product-engineering/engineering/code-graph/code-insights/processes.md
```diff
@@ -59,6 +59,24 @@ Our current process is as follows:
...
+
+### Time commitment
+
+Being a project lead takes a bit more time than just working on a single issue. If you are a project lead then your capacity will be reduced. Project leads should make sure to take this into consideration when planning other iteration work.
+
+### Responsibilities
+
+1. Have a full understanding of the requirements and the work that needs to be done across the frontend and backend. The main benefit here is catching any misunderstandings or integration issues before they happen.
```
Joelkw, line 81:
```suggestion
1. Be the first point of engineering contact for product and design
1. Have a full understanding of the requirements and the work that needs to be done across the frontend and backend. The main benefit here is catching any misunderstandings or integration issues before they happen.
```

Joelkw (approved):
Thank you!

leonore (approved):
general responsibilities of a project lead lgtm, I think it's better if we get this in now and then iterate on it if needed

content/departments/product-engineering/engineering/code-graph/code-insights/processes.md
```diff
@@ -59,6 +59,25 @@ Our current process is as follows:
 
 - If one or more issues that were planned for an iteration are looking to **not get finished** (which includes testing) in the [current iteration](https://github.com/orgs/sourcegraph/projects/200/views/1) (while maintaining sustainable work practices) the assignee **raises this as soon as possible asynchronously** to the team (including the PM and EM), e.g. on the GitHub issue or Slack. These issues then **become _proposed_ issues for the next iteration** (meaning nothing carries over automatically, but we also don't just drop and forget missed issues).
 
+## Project Leads
+
+Larger issues, especially ones that involve multiple domains, may benefit from a "project lead". A project lead assumes full responsibility for gathering all requirements to complete an issue. The project lead is expected to have full understanding of the issue and pull in any resources or team members needed to complete the task.
```
unclejustin:
```suggestion
Larger projects, especially ones that involve multiple domains, may benefit from a "project lead". A project lead assumes full responsibility for gathering all requirements to complete a project. The project lead is expected to have full understanding of the issue and pull in any resources or team members needed to complete the task.
```

unclejustin:
Excellent suggestions everyone! Merging 🎉 

@chwarwick project leads are expected to be engineers.

PR #2709, opened 2022-03-15
Update headcount planning documentation

nicksnyder:
This update proposes a clarification the process for the rest of FY23.

PR #2991, opened 2022-04-06
Add note about CS headcount

nicksnyder:
From: https://github.com/sourcegraph/handbook/pull/2709#issuecomment-1074877016


virginiaulrich (approved):
Thank you! I am sorry I couldn't figure out where to capture this ... now this feels really obvious.

PR #3021, opened 2022-04-08
Adds End-to-End review as onboarding step

danielmarquespt:
Adds end-to-end review as a possible onboarding step, with a brief description on how to do it.

Moves the Manager checklist to the bottom, since there are more Designers than Design Managers, the design manager section is not as useful for most of the readers of this page


rrhyne (approved):
Great change!

rrhyne:
Merging this so it's there for Paulo.

PR #3067, opened 2022-04-13
Add process for organization deletion

pjlast:


content/departments/product-engineering/engineering/process/cloud_processes.md
```diff
@@ -0,0 +1,43 @@
...
+DELETE * FROM settings WHERE org_id = $ORG_ID;
+```
+
+```sql
+DELETE * FROM orgs WHERE id = $ORG_ID;
+```
+
+### Other useful queries
+
+A list of useful DB queries can be found in this Google Sheet: https://docs.google.com/spreadsheets/d/1Z1-7uJwtF2etZFeqTcJS4z9WcAjKlkCxt8HQ931D3dA
```
miveronese, line 43: @pjlast , apologies in advance if I am missing some context here... I added a comment on slack about this. My suggestion/idea is to, instead of asking the users to run DB queries, instruct them to use the Delete Organization area (on Organizations > Settings) in the UI + explain that the organization and all the resources associated to it will be hard deleted.   What do you think?

miveronese (approved):
All looks good to me!

content/departments/product-engineering/engineering/process/cloud_processes.md
```diff
@@ -0,0 +1,43 @@
+# Sourcegraph Cloud Processes
+
+## Closing down an organization on Sourcegraph Cloud
+
+During the open-beta, customers receive 30 days to try out Sourcegraph Cloud with Organizations. Once this period has passed, customers need to reach out in order to continue leveraging Sourcegraph Cloud.
+
+The following process documents the queries that need to be executed in order to completely remove an organization if a customer does not reach out.
+
+### Queries
```
unknwon, line 9: Hmm, don't we just need to hit the org hard delete GraphQL endpoint and the no need to have direct access to the DB? (since this section it basically describing what the hard delete endpoint does exactly in my understanding).

content/departments/product-engineering/engineering/process/cloud_processes.md
unknwon:
```suggestion
A list of useful DB queries can be found in [useful DB queries]( https://docs.google.com/spreadsheets/d/1Z1-7uJwtF2etZFeqTcJS4z9WcAjKlkCxt8HQ931D3dA).
```

content/departments/product-engineering/engineering/process/cloud_processes.md
pjlast: @unknwon from what I understand this is for support which often does things with access to the DB. We might have to speak to @ryphil when he's back just to make sure what it is he had in mind

unknwon: > @unknwon from what I understand this is for support which often does things with access to the DB. We might have to speak to @ryphil when he's back just to make sure what it is he had in mind

If we have no other ways, I'd agree we do need access to DB, but if we have equivalent and less error-prone ways, I'd suggest avoid direct access to DB 😁

ryphil: @pjlast or @unknwon - I'd also suggest avoiding direct DB access. 

Regardless, given then news that we won't be onboarding many new teams, we shouldn't need to run this process very often. And rather than hand off to support, our team would handle the request. Thus, i'd say whatever is easiest to finish this task up quickly.

PR #3087, opened 2022-04-14
Add FY23 Q2 focus to strategy

christinaforney:


malomarrec:
Awesome to have this in the handbook!

content/departments/product-engineering/strategy-goals/index.md
```diff
@@ -55,7 +57,24 @@ One difference between this year and last is that we anticipate more competition
...
+
+**We need to improve our deployments to meet customer demand**
+We have a great platform and our customers are struggling with getting to a point where they can use it through a trial or production instance in a reasonable amount of time. We need to ensure that deployments are as seamless as possible and don’t get in the way of customer value. In addition, we continue to see increased interest in Sourcegraph managed deployments since they provide a way to avoid the burden of self-hosting and maintaining Sourcegraph, so we need to quickly scale our ability to support more managed instances to meet these demands. From our customer perspective, managed instances are a Cloud SaaS offering of our product and we also need to think of it in this way. Over time (~2 years), we expect automated managed instances to evolve into a streamlined Sourcegraph hosted self-service solution.
```
rafax: Would it make sense to drop ```We have a great platform``` in this context - the latter part of the sentence suggests customers are struggling to use it?

christinaforney:
```suggestion
We have a great product and our customers are struggling with getting to a point where they can use it through a trial or production instance in a reasonable amount of time. We need to ensure that deployments are as seamless as possible and don’t get in the way of customer value. In addition, we continue to see increased interest in Sourcegraph managed deployments since they provide a way to avoid the burden of self-hosting and maintaining Sourcegraph, so we need to quickly scale our ability to support more managed instances to meet these demands. From our customer perspective, managed instances are a Cloud SaaS offering of our product and we also need to think of it in this way. Over time (~2 years), we expect automated managed instances to evolve into a streamlined Sourcegraph hosted self-service solution.
```

Updated to distinguish the product, but that the pain point is around the setup/configuration.

PR #3132, opened 2022-04-20
add customer to the MI upgrade list

mucles:


daxmc99 (approved):
LGTM, 
TIL about this process.

daxmc99:
See also this comment: https://github.com/sourcegraph/customer/issues/820 
Looks like we might be done after this

PR #3216, opened 2022-04-26
Remove terminated teammates

CecilyBlack:


varungandhi-src:
The problem with the PR check is that the reporting relationship is not well-formed. For example, various people (such as Fabiana) are marked as reporting to `director_demand_generation`. However, this PR removes the entry for Sarah Reece, who is marked to have the `manager_role_slug: director_demand_generation`. This means that it's not clear how to connect Sarah's previous reports to the org chart. The right fix here would be to fix the `reports_to` field for Fabiana and others to point to the right manager.

A similar problem applies to some other situations as well, such as Luke Taylor being marked as reporting to `head_international_sales`, however, this PR removes Prakash Durgani who had that as the `manager_role_slug`.

CecilyBlack:
Great, thank you so much Varun! I will find out the reporting structures
for these teams and then make the edits once I have that information :)

PR #3238, opened 2022-04-28
Add Extend MI to upgrade list

filiphaftek:
Extend MI added to upgrade list.


sanderginn:
Maybe we can generate this list based on the contents of the MI repo? Seems like this can become a bit of a risk when the number of MIs will ramp up

sanderginn (approved):

filiphaftek:
> Maybe we can generate this list based on the contents of the MI repo? Seems like this can become a bit of a risk when the number of MIs will ramp up

Great idea :) @mucles what do you think?

mucles:
> > Maybe we can generate this list based on the contents of the MI repo? Seems like this can become a bit of a risk when the number of MIs will ramp up
> 
> Great idea :) @mucles what do you think?
+1 to that, I think this will be exponentially harder to maintain and keep track of as the number of MIs ramp-up

mucles (approved):

mucles:
We would need to consider how to map the related customer issue with each corresponding content in that repo to automate the generation of that list 👍

sanderginn:
I [asked if customers can get an MI label](https://sourcegraph.slack.com/archives/C01EC12SB33/p1651234299047249?thread_ts=1651231990.334809&cid=C01EC12SB33) where necessary. That should allow us to just generate the list based on the labels attached to customer issues, and we don't have to concern ourselves with mapping to the repo. Added benefit is that it would expose outdated customer info in the accounts repository :)

filiphaftek:
@sanderginn I am merging this not to omit Extend in next upgrade round. We can open new PR when labels are in-place for MI.

sanderginn:
👍 sounds good

PR #3316, opened 2022-05-04
update product gap opportunity guidance

Joelkw:
PMs want to use product gaps as accurate reflections of feedback that impacts opportunities, but if we tag unrelated-to-expansion-products feedback to expansion-products Opportunities, this renders the signal noisy, and over time, we'll stop using it correctly or trusting it. [See this discussion](https://sourcegraph.slack.com/archives/C0W2E592M/p1650500782558299?thread_ts=1650494933.548379&cid=C0W2E592M) for more context.


Joelkw:
@ZrnH going to go ahead and merge this!

PR #3342, opened 2022-05-05
Link to new issue template

rebeccadee:
Linking to fixed issue template added in https://github.com/sourcegraph/about/pull/5333


rebeccadee:
@csbailey5t Just seen I don't need approval to merge so will go ahead and do that!

PR #3348, opened 2022-05-05
Add delivery pairing exercise details

caugustus-sourcegraph:
Add instructions for both candidates and interviewers for the delivery-specific pairing exercise.


kevinwojo (commented):
LGTM

PR #3374, opened 2022-05-06
Add FY23Q2 OKRs for the IAM team

ryphil:


content/strategy-goals/strategy/cloud/iam-admin/index.md
```diff
@@ -1,5 +1,47 @@
...
+
+- Key Result 1: Identify and deliver an improved usage analytics experience for admins.
+
+**Product and Engineering OKR Alignment:** Objective #1: Faster Time-to-Value by enterprises. Today, enterprise admins have very little visibility into usage within their organization. They are forced to rely on the basic stats provided by Sourcegraph or anecdotal reviews from their team. Introducing more fine grained analytics will enable admins to easily quantify value to their team and aid Sourcegraph AE/CE in driving upsell conversations based on usage within the org.
+
+### Problem Area 4
+
+Today, Sourcegraph does not provide administrators the ability to specify who can and cannot sign up for Sourcegraph, yet this functionality is critical for customers who want to adopt a land-and-expand strategy for adoption within their organization.
+
+- Key Result 1: MeLi and Capital One product deployment is unblocked leveraging a short-term hack
```
bcreager: Do we use customer names in the handbook?

content/strategy-goals/strategy/cloud/iam-admin/index.md
```diff
@@ -1,5 +1,47 @@
 # IAM and Admin Experience Team
 
-## Current and Future Goals
+This page outlines the short-term strategy for the IAM and Admin Experience Team. You can find the overview of Sourcegraph Product and Engineering FY23 Q2 focus [here](../../../../departments/product-engineering/strategy-goals/index.md#fy23-q2-focus)
 
-This team is newly formed and is currently defining goals.
+## Q2 Objective and Key Results
+
+### Problem Area 1
```
bcreager: Using numbers here makes is feel like they are in priority order, which I don't think is the intent.  Could we name the problem areas or maybe just make a bulleted list so we don't infer priority?

ryphil: Agreed - I removed the number and replaced with a short title of the problem area. How does this feel?

bcreager (approved):

quinnkeast (approved):

content/strategy-goals/strategy/cloud/iam-admin/index.md
```diff
@@ -1,5 +1,47 @@
 # IAM and Admin Experience Team
 
-## Current and Future Goals
+This page outlines the short-term strategy for the IAM and Admin Experience Team. You can find the overview of Sourcegraph Product and Engineering FY23 Q2 focus [here](../../../../departments/product-engineering/strategy-goals/index.md#fy23-q2-focus)
 
-This team is newly formed and is currently defining goals.
+## Q2 Objective and Key Results
+
+### Problem Area: Align on Team Vision
```
unknwon, line 7: qq: These are sub sections of "Objective and Key Results" but why is they are named as "Problem Area:" instead of "Objective:"?

content/strategy-goals/strategy/cloud/iam-admin/index.md
```diff
@@ -1,5 +1,47 @@
...
+
+### Problem Area: Improve Admin Analytics
+
+Today, Sourcegraph Admins do not have a clear understanding of how their team is using (or not using) Sourcegraph. This information is valuable to help administrators (1) convey the value of Sourcegraph to executives and (2) make key decisions required to do their job effectively, such as monitoring remaining licenses.
+
+- Key Result 1: Identify and deliver an improved usage analytics experience for admins.
+
+**Product and Engineering OKR Alignment:** Objective #1: Faster Time-to-Value by enterprises. Today, enterprise admins have very little visibility into usage within their organization. They are forced to rely on the basic stats provided by Sourcegraph or anecdotal reviews from their team. Introducing more fine grained analytics will enable admins to easily quantify value to their team and aid Sourcegraph AE/CE in driving upsell conversations based on usage within the org.
+
+### Problem Area: Allow admins to control sign-ups via externally defined groups
```
unknwon: nit: other "Problem Area" use Title case but this one is using sentence case (we should use the latter according to our handbook, but being consistent is just as good enough).

content/strategy-goals/strategy/cloud/iam-admin/index.md
```diff
@@ -1,5 +1,47 @@
...
+
+Today, Sourcegraph does not provide administrators the ability to specify who can and cannot sign up for Sourcegraph, yet this functionality is critical for customers who want to adopt a land-and-expand strategy for adoption within their organization.
+
+- Key Result 1: Unblock deployment for blocked customers
+- Key Result 2: A technical roadmap is written to provide admins the ability to limit user sign-up based on an externally defined entity (i.e. a GitHub Organization or Team, Gitlab organization, LDAP group for SAML connections).
+- Key Result 3: Short term improvements are implemented to provide admins the ability to limit user sign up based on externally defined entities (i.e. GitHub organization, GitLab organization, IDAP group)
```
unknwon, line 45: Does the order/number of the key results matter? Looking at other key results their order seems making sense to me in the order they are define. But the order here which would make sense to me is 3, 1, 2.

content/strategy-goals/strategy/cloud/iam-admin/index.md
ryphil: Fixed in next commit!

content/strategy-goals/strategy/cloud/iam-admin/index.md
ryphil: @unknwon - The ordering is closer to "How it happened", rather than any amount of importance. The #1 is already done and was *very* hacky solution because these customers were blocked on a production deployment. Then we started to build the plan for #2, and realized there are a few other short term fixes that would make our exiting solution less hacky, while still providing short term value to customers wanting this solution.

 So ultimately, the goal in Q2 is to provide a few short term fixes for customers and gather agreement on a long-term product improvement.

content/strategy-goals/strategy/cloud/iam-admin/index.md
ryphil: @unknwon - Good question and callout. 

There are a few reasons: 
- Since we are fairly new, the various areas are at different levels of scoping. So we felt defining as problem areas was a good normalizer for areas we do not have any outcomes defined yet. 
- It brings the user voice problem into the planning process so we make sure our plans for the quarter are tied to user problems
- It lets our team be in the driver seat of defining the best solution, instead of embedding a solution in results.

ryphil: @unknwon - I'm going to merge this and we can update if you feel we should adjust the heading.

content/strategy-goals/strategy/cloud/iam-admin/index.md
unknwon: Makes sense, thanks!

content/strategy-goals/strategy/cloud/iam-admin/index.md
unknwon: @ryphil thanks, no need to update the headings as was literally a question :)

Though just speaking on the matter, "solving problem area XYZ" could just be our objectives, if that makes sense to be an "objective" at all.

PR #3401, opened 2022-05-10
added bio

JeanPaulSG:
added my bio to the handbook team page


JeanPaulSG:
ah, I used an incorrect role slug to refer to Jeff. Should be head_of_technical_writing

PR #3405, opened 2022-05-11
add DPA FAQ

tammy-zhu:


collettesg (approved):
I think this is fine as-is & have provided my approval. You may want to consider whether we should add a note clarifying that there is a different process for suppliers (or cross-linking to that page), if the context of the handbook page where these FAQs live can cause confusion (like if someone were to search the handbook for "DPA" and go here - will they understand that this doesn't apply to all contracts?)

PR #3419, opened 2022-05-11
add new instance to upgrade_managed_issue_template.md

mucles:


michaellzc:
I think PR like this in the future we can just merge without review

michaellzc (approved):

PR #3420, opened 2022-05-11
bugfix/ MD date hover function

### LINKED ISSUE #2520: Date tooltips think fiscal years start in August
This could be misleading because the OKRs page is linked from the onboarding checklist, and a new hire doesn't know yet how fiscal years work:

image

Code: https://sourcegraph.com/github.com/sourcegraph/handbook/-/blob/src/lib/dateHighlighter.ts

st0nebreaker:
Fixes bug where markdown fiscal year and quarter dates were inaccurate, namely Q1 & FY in ss's below, by adding in logic to account for these in `parseFiscalInterval()`.

Closes #2520

### BEFORE:
**Quarter bug:**
image
**Fiscal year bug:**
image
### AFTER:
**Quarter fixed:**
Screen Shot 2022-05-11 at 2 30 29 PM
**Fiscal year fixed:**
Screen Shot 2022-05-11 at 2 30 36 PM


src/lib/rehypeMarkupDates.ts
```diff
@@ -123,8 +123,18 @@ function parseFiscalInterval(input: string): { start: string; end: string } | nu
...
+    let startYear
+
+    if (input.includes('Q4') || !input.includes('Q')) {
+        endYear = parseInt(startFiscalYearString, 10)
+        startYear = endYear - 1
+    } else {
+        // start & end year are the same if evaluating any Q that isn't Q4
+        endYear = parseInt(startFiscalYearString, 10) - 1
+        startYear = endYear
+    }
```
felixfbecker: **issue**: This is incorrect for strings referring to the second half of a fiscal year, like `FY22H2`. For example here: https://handbook.sourcegraph.com/strategy-goals/strategy/code-graph/code-insights/#strategy-for-fy22h2
The second half of the fiscal year ends in the next calendar year after its start.

Taking a step back, this code looks again at the `input` string – but at this point in the code, we've already parsed out `startFiscalYearString`, calculated `startMonth` and calculated `durationMonths` 🤔 

With that available, I think we don't even need an `if` statement and look at `input` anymore, we can just use `startFiscalYearString` and mathematically calculate the end year:

```suggestion
    const startYear = parseInt(startFiscalYearString, 10) - 1
    const endYear = Math.floor(startYear + (startMonth + durationMonths) / 12)
```

Example calculations:

```ts
// FY22H2
const startMonth = 6
const durationMonths = 6
const startYear = 2022
const endYear = Math.floor(2022 + (6 + 6) / 12) // 2023

// FY22Q4
const startMonth = 11
const durationMonths = 3
const startYear = 2022
const endYear = Math.floor(2022 + (11 + 3) / 12) // 2023

// FY22Q1
const startMonth = 1
const durationMonths = 3
const startYear = 2022
const endYear = Math.floor(2022 + (1 + 3) / 12) // 2022
```

Did I overlook anything?

st0nebreaker: Just getting back to this, I didn't register the `FY22H2` scenario. Thanks for pointing it out, and that's a great simpler suggestion, I put it in [dbf1e59](https://github.com/sourcegraph/handbook/pull/3420/commits/dbf1e5921a31c3590e209b8261533d06d991d253). Let me know if you see any gaps with this

felixfbecker (approved):

unknwon:
Thank you!

PR #3447, opened 2022-05-12
update ci playbook

davejrt:
removes references to the old autoscaler, and rewrite some references to state, which are now redundant with stateless agents.


content/departments/product-engineering/engineering/process/incidents/playbooks/ci.md
```diff
@@ -217,22 +204,6 @@ In order to handle problems with the CI, the following elements are necessary:
      1. **Yes**, that's a flake. See the [Spotted a flake scenario](#spotted-a-flake)
    - **No**: it's not a flake, reach out the team owning those tests.
 
-### Restarting the agents
```
bobheadxi, line 220: Doesn't have to be this PR, but you might still want to restart agents (e.g. if a dispatch is borked) by deleting an entire dispatched Kubernetes job

bobheadxi (approved):
Thanks for combing through this and bringing it up to date! ❤️

content/departments/product-engineering/engineering/process/incidents/playbooks/ci.md
davejrt: yeah, so in typical fashion I was going to leave some comments here asking for feedback...but pushed and then went off on something else. 

So, my gut feel is we want to probably want to reference deleting a job entirely if a series of agents that have been deployed is failing. 

I'll follow up here.

jhchabran (approved):

PR #3451, opened 2022-05-12
Retain cloud events indefinitely

ebrodymoore:
Proposing to retain these events indefinitely - let me know if anybody has an issue with this! No new tracking, and no real reason we did this in the first place.


ebrodymoore:
Talked with @collettesg over email - better to have a limit on here so setting at 10 years

PR #3466, opened 2022-05-14
Remove a responsibility from CEO role description

gronk-sg:
Given @amenne has been appointed as VP Customer Engineering several months ago it seems managing the CE team can be removed from @sqs CEO responsibilities.


RoudyBob:
Seems reasonable. LGTM.

PR #3513, opened 2022-05-18
Add specifics around Covid-19 exposure protocols

cassiemelani:


cassiemelani:
Hi team, due to more instances arising, I wanted to carve out our policies more so people know what to expect before, during, and after traveling if exposed to COVID. 

Please review by EoW! Thank you :)

PR #3533, opened 2022-05-20
Updating the Vendor Contract section

doragrgic:
I have updated the vendor contract section to point to our new Vendor Request Process handbook page.


tammy-zhu (approved):

tammy-zhu (changes requested):
Hi @doragrgic one question here:  I see that the Vendor Request Process applies to only vendor contracts that "touches Sourcegraph data and/or client’s data." What about for those that don't? Should we still keep the existing vendor request process for those that don't touch sourcegraph data or client's data?

Also, what if a vendor contract touches teammate data but no "sourcegraph data" or "client's data"?

doragrgic:
@tammy-zhu Sourcegraph's data should include our personal data as we are employees of the company. Happy to change the wording to explicitly mention personal data as well. We should use the same process for all vendors.

PR #3558, opened 2022-05-23
search core: Describe the Zoekt deployment process.

mpimenov:


content/departments/product-engineering/engineering/code-graph/search/core.md
```diff
@@ -83,6 +83,14 @@ Our team is geographically and timezone diverse. The handbook has a [large page
 - For support rotation picking up incidental work. Tasks in support backlog to capture this, more intentional work around support and papercuts versus trying to work normally and getting interrupted.
 - Focus on more feature flag driven development to simplify on-premise changes and speed up MTTR for Cloud.
 
+### Releases
+
+Most of the work related to releasing our code fits the general description of [Engineering processes](../../process).
+
+However, there is an important thing to mention. Since Zoekt lives in a separate repo, we need to periodically pull changes from the Zoekt repo to the main Sourcegraph repo. This can be done either manually by altering the go.mod and go.sum files in the main repo, or by merging the corresponding PR opened by [the bot](https://github.com/sourcegraph/sourcegraph/pulls?q=is%3Apr+author%3Asourcegraph-buildkite+zoekt).
```
stefanhengl: There is also the script `https://github.com/sourcegraph/sourcegraph/blob/main/dev/zoekt/update` which I usually use.

content/departments/product-engineering/engineering/code-graph/search/core.md
```diff
@@ -83,6 +83,14 @@ Our team is geographically and timezone diverse. The handbook has a [large page
...
+
+This is good enough to see our changes in an upcoming release, but if you want to see them immediately at Sourcegraph Cloud, another step is needed. While Sourcegraph Cloud is continuously deployed from the main repo code, we stall some parts of it by providing an intentionally wrong version number in the deployment script. The reason is these parts are too large and Sourcegraph Cloud may temporarily go out of service for a small but noticeable time while they are restarting. Zoekt is one of such parts (gitserver is another). For information on how to bump the Zoekt images and deploy the newer versions, refer to [Continuous Deployment Process](../../process/deployments/index.md#continuous-deployment-process) and [Manually deploying a service to sourcegraph.com](../../process/deployments/playbooks.md#manually-deploying-a-service-to-sourcegraphcom).
```
stefanhengl: I believe the reason why we provide an intentionally wrong version is because some time ago the delay between merging an update of Zoekt to Sourcegraph and the start of the rollout to Cloud was very long. By updating the sha manually in the deploy repo we trigger a rollout immediately.

Meaning, if we moved back to automatic deployments, we wouldn't get more rollouts, but the rollouts might not happen when we expect. We should confirm whether this is still the case because a lot of things have changed since then.

stefanhengl (approved):

content/departments/product-engineering/engineering/code-graph/search/core.md
stefanhengl: Should we also mention pre-prod here?

content/departments/product-engineering/engineering/code-graph/search/core.md
mpimenov: done

content/departments/product-engineering/engineering/code-graph/search/core.md
mpimenov: I've added the other reason for now. I like the idea of checking and leaving only one if the other turns out wrong.

Re pre-prod: I don't think so, it's mentioned if one follows the link.

mpimenov: > By updating the sha manually in the deploy repo we trigger a rollout immediately.

I'm leaving this version you suggested and removing the unconfirmed one from the first version of the PR.

PR #3626, opened 2022-05-26
Updates MI creation steps to reflect `mg` commands

sanderginn:


daxmc99 (approved):
I believe we have a `1.1` folder where we are making the updates for the new process which we can switch in the future.

daxmc99 (changes requested):
Let's agree the 1.1 section is complete before merging this

sanderginn:
OK - should we stop using the v1.1 process that we have so far, then? We've used it on at least 1 new MI. Do you think it's necessary to revert the creation instructions to the last state of v1.0 (and add v1.1 instructions in the appropriate folder)?

sanderginn:
I fudged the rebase 🤦‍♂️ 

The original creation instructions are back the pre v1.1 version, and the new instructions are filed under the 1.1 directory.

PR #3645, opened 2022-05-27
Add note that incidents are created by people

daxmc99:
Incident creation process had an implicit assumption  of a human being involved that needs to be made explicit

### DIFF
--- a/content/departments/product-engineering/engineering/process/incidents/index.md
+++ b/content/departments/product-engineering/engineering/process/incidents/index.md
@@ -15,6 +15,9 @@ Additionally, on big announcement days (funding, product launch, campaign launch
 
 All incidents are announced in the #incidents channel automatically through incident.<unlink></unlink>io, and past incidents are available in the [incidents.io dashboard](https://app.incident.io/incidents) and our [past incident postmortems](#past-incident-post-mortems).
 
+All incidents are opened by a team member and closed by a team member. We currently have no automated process for incident creation
+or closing. This is a deliberate decision such a that a human must declare an incident and must declare it resolved.
+
 ## Process
 
 ### Identification


danieldides (approved):
I like this. Should we also add a requirement that the human who opens the incident provides the initial context (either in the thread directly or a link to something preexisting)?

bobheadxi (changes requested):
Copied from [my comment in RFC 592](https://docs.google.com/document/d/13kVYJSoTWYgyeBFXeJnasF3WU7SHyOjhSNpp9yV_pvY/edit?disco=AAAAZwrIZfo):

> I'm inclined to agree initially [that we might not want to automate incident creation], for the same reason I argued against automating incident creation for branch locks. That said, breakages in the deployment smoketests should never happen, and I think it is appropriate to create an incident here. Opsgenie [in the deployment smoketest scenario] is difficult because we cannot really assign a specific engineer - a deployment could contain any number of changes, many of which might not be relevant at all, unless the Cloud DevOps team would like to take on the Opsgenie assignment. The advantage of an incident is that it leaves the door open to participation

daxmc99:
@bobheadxi For me, this PR is a _not_ a change. This was the default prior to the work in RFC 592.  Your comments in that RFC are valid and I think we should have a larger discussion there. But after https://github.com/sourcegraph/deploy-sourcegraph-cloud/pull/16663 this is current state. 

This could be changed tomorrow but the above change is what is true currently, correct?

bobheadxi:
@daxmc99  I'm not sure it's a default yet - my understanding was that outside the build incident case where a specific case was made against automated incidents, the _only_ reason we didn't have automated incidents is because of the lack of API (it only became available last month: https://github.com/sourcegraph/sourcegraph/issues/31854#issuecomment-1113175300), not because we had a conscious decision that all incidents were created by humans 🤔 

Basically, I'm of the opinion that we do not have a stance in regards to incident creation, and the discussion should be in regards to introducing one now that the API is available and incident automation is an option at all

daxmc99:
> Basically, I'm of the opinion that we do not have a stance in regards to incident creation, and the discussion should be in regards to introducing one now that the API is available and incident automation is an option at all

Okay, I think we have a difference of opinion here. I think the conclusion of https://github.com/sourcegraph/sourcegraph/issues/31854 we should update this handbook page with the result

PR #3708, opened 2022-06-01
Document current stewardship of package repos

malomarrec:
Package repos are an important part in building the experience we're planning for code security. Also see this MVP tracking issue: https://github.com/sourcegraph/sourcegraph/issues/37012.
I think we should document an owner/stewart, and it sounded like this should be code intel?


macraig:
@malomarrec when we say package repos are we referring to the entity? In that case would it be better if it was owned by the same team that owns the repo entity (if there is such a team, probably not)?

I'm thinking code intel is a good owner for everything related to making precise packages available but not sure about the package as an entity scope (might be search core, repo management if we treat them as repos?) I'm still not sure where this should fall. cc @olafurpg who might want to chime in

olafurpg:
I am increasingly leaning towards recommending that Code Intel takes over ownership of packages. I estimate that it will not require too much effort to implement the recommendations from RFC 698, which is currently the biggest unowned workstream related to packages. Our team has the most domain expertise for packages, and we also have a strong incentive to make them work well.

> In that case would it be better if it was owned by the same team that owns the repo entity

By going this route I am concerned we end up in a state where packages stop evolving. Although we encode packages as repos, most of the remaining work is adding functionality that's different from other code host integrations (frontend work to make packages a first-class entity, and implement syncing strategies that are unique to package hosts).

malomarrec:
@macraig I think someone needs to own the end-to-end experience ("what it means to search packages"), because just treating packages as standard repos does not create a great UX, as explained in [RFC 698 WIP: Packages should be a first class entity](https://docs.google.com/document/d/1aGTzOKXJkf29FXUaDb3uevmIXmX09hYS11UKSp6j2SA/edit#heading=h.eklszdrbn53c).

We could of course have a team owning the experience and another the infrastructure (like search and search core), if it's not possible to have a single owner. And to be clear by "own" I don't mean "build everything".

mrnugget:
> I am increasingly leaning towards recommending that Code Intel takes over ownership of packages. I estimate that it will not require too much effort to implement the recommendations from RFC 698, which is currently the biggest unowned workstream related to packages. Our team has the most domain expertise for packages, and we also have a strong incentive to make them work well

Hmm, does Code Intel have enough resources with the right expertise to implement that?

In [your comment here](https://docs.google.com/document/d/1aGTzOKXJkf29FXUaDb3uevmIXmX09hYS11UKSp6j2SA/edit#heading=h.o8zqys3ugzcm) you wrote that you don't have the resources (because of indexer work) and not the right expertise. I know that Eric Fritz and Cesar and Noah, for example, are super busy with implementing RFC 619, plus executor work, plus SCIP support in backend, etc.

Does that mean the Language Tools subteam will take this on? If so, is "area of expertise" still a concern? (We're talking about possibly deep platform work (to fix syncing of packages, for example)).

olafurpg:
@mrnugget All of your concerns are valid and it's unlikely the language tools team will take on work outside of our direct needs, which primarily involves implementing RFC 698 and adding support for new ecosystems like Rust Crates. It would not include implementing novel syncing strategies for dependency search. It's better if we find an owner for all package-related projects but nobody has volunteered for that yet.

olafurpg:
Instead of going back-and-forth about assigning a owner for everything, it might be more productive to divide the work between multiple owners.

malomarrec:
@olafurpg @scalabilitysolved do you have a proposed split in mind?
We could document, for example, that Rob and a PM own the experience, code intel own x, etc. But we'd need to make that split clear.

olafurpg:
@malomarrec I am concerned that it will become quickly outdated if we write down which exact sub-components are owned by which teams (or people). It might be simpler administratively to list multiple owners for "Packages" and create a `#packages` Slack channel where anyone can ask questions about packages.

mrnugget:
I don't think a split ownership helps to answer the original question: "who's going to take care of package repos, fix the operational concerns (rate limiting, overloading code hosts), implement RFC 698 properly so that package repos are usable from a [dependency] search perspective?" If we define a split ownership now, we're describing the status quo.

What I want to avoid is that we meet back here in 2 months, after RFC 698 is implemented and another problem popped up.

I want someone to own package repos long-term and with "own" I mean: fix problems that come up, take care of operational concerns, be able to answer questions that come up in Slack (like Alex' questions around rate limiter in past few weeks), assist others in using package repos in their own projects, implementing new features, etc.

I really don't care _who_ owns it, but that it's owned in this way. If Code Intel/Olaf want to own it and they also understand the same thing under ownership, then that's fine with me.

scalabilitysolved:
An open question for me here would be, what's the near term future of the Code Security team going to look like @mrnugget @malomarrec ?  Is this team likely to dissolve? Will your focus change? (Because @mrnugget has been doing a lot of work in this space already).

malomarrec:
@scalabilitysolved there is a little uncertainty around this. I think it will be clearer in 2 days. I'll plan to get back to this then.

olafurpg:
@mrnugget I want the same things but I don't think it's realistic that we'll have a dedicated owner to address all package-related issues. For the foreseeable future, it's not unreasonable to have shared ownership and tackle issues on a case-by-case basis by discussing via a `#packages` Slack channel or individual GitHub issues.

We've been talking about ownership for packages over several weeks now and still don't have any volunteers. I'd much rather discuss the actual underlying issues. For example, rate limits should be solvable if we stop sending repeated requests for already downloaded packages.

mrnugget:
Hmm, not sure I'm following. Does that mean you/Code Intel won't be able to take over ownership as I described it in my last comment? But I assume you still want to keep working on package repos, right? To add support for the Rust crates?

My worry right now is that if we continue like this - no dedicated owner that owns package repos in the way I sketched out above, new stuff built on top - we don't improve the situation and might make things worse.

> I'd much rather discuss the actual underlying issues. For example, rate limits should be solvable if we stop sending repeated requests for already downloaded packages.

From my perspective that is what we've been doing - in Slack, in the RFC, in Zoom calls. But it doesn't lead to problems being fixed. Rate limiting is a good example here: somebody needs to take that on, but nobody felt responsible to do so. Yesterday another issue with rate limiting popped up. I dug into it with Erik to find out why it's happening, but I also can't dedicate time right now to fix these issues. So we've talked about these problems but we haven't found someone to take them on.

---

The bigger problem here is that this discussion would be much clearer if I'd know what the future of Tiger Team looks like after Q2. I think that's what @malomarrec is currently trying to find out.

olafurpg:
>  Does that mean you/Code Intel won't be able to take over ownership as I described it in my last comment? But I assume you still want to keep working on package repos, right? To add support for the Rust crates?

My recommendation is that Code Intel takes ownership of fixing blocking issues for precise intel and other Code Intel products. This includes working on crates.io but it does not include UX issues around dependency search.

> From my perspective that is what we've been doing - in Slack, in the RFC, in Zoom calls. But it doesn't lead to problems being fixed. Rate limiting is a good example here: somebody needs to take that on, but nobody felt responsible to do so. 

If dependency search is blocked by rate limiting then I think it would be natural for the Code Security team to take ownership of fixing that issue. Rate limiting has not come up as a blocking issue for Code Intel, which is the reason that nobody from our side has volunteered to fix the issue. I'm happy to consult on the issue, based on the low number of package+version combinations that we are fetching I don't think there's an inherent/fundamental issue that can't be fixed or worked around.

Same for the syncing strategy when trigging a dependency search. If it's confusing for users that the search results are incomplete then the Code Security team would take ownership to solve that issue by either

* displaying in the UI that the results are incomplete, or
* displaying a progress bar in the UI to hint that a background process is handling the query, or
* implementing a separate syncing strategy so the query can be immediately resolved with complete/comprehensive results.

mrnugget:
I'd personally prefer to have clear long-term ownership vs. multiple teams working on issues that affect them, since I think package repos and dependency search would benefit from a dedicated owner (product & engineering).

But it looks like there's too many variables right know that make it hard to decide on that. And I fully acknowledge that the biggest unknown right now is "what happens to the tiger team after q2?"

So how about this: we leave it undecided and go with the approach you sketched out, @olafurpg (everyone working on issues/blockers that affect them), and then once we know more (might be in a few days!) we revisit.

rrhyne:
> @malomarrec when we say package repos are we referring to the entity? In that case would it be better if it was owned by the same team that owns the repo entity (if there is such a team, probably not)?
> 
> I'm thinking code intel is a good owner for everything related to making precise packages available but not sure about the package as an entity scope (might be search core, repo management if we treat them as repos?) I'm still not sure where this should fall. cc @olafurpg who might want to chime in

@macraig no team owns the repository page as far as I can tell. The team focused on repos is supporting high value work for some of our largest clients and wouldn't be able to take that on.  

Personally, I think that because code intel released package repos, it should be responsible for getting the feature to a state where they are usable and production ready. @olafurpg and I worked closely on changes that need to occur to make this happen in RFC 698. I can support those efforts from the design side.

macraig:
@rrhyne I agree code intel should own work related to package repos and precise code navigation. We're also owning the changes you and @olafurpg synced on from RFC 698 as it ties into part of our precise code intel scope. The part which is still unclear to me is who will own and drive the vision for packages as a first-class entity overall. The code security team was interested in driving it (cc @malomarrec in case this has changed) but if the team doesn't continue to exist in Q3 then we once again remain without a high-level long-term owner.

PR #3710, opened 2022-06-01
all: replace links to sourcegraph master to main

keegancsmith:
Spotted a place linking to the master branch. We may end up one day deleting the branch, so lets avoid potentially stale links spreading. Ended up updating all references I could find with a simple regex:

```shell
ruplacer --go '(sourcegraph/sourcegraph/[^/]+)/master/' '$1/main/'
```


jeanduplessis:
Thanks @keegancsmith 🙏

PR #3730, opened 2022-06-01
Merge CE and CS handbook sections

jeanduplessis:


amenne (approved):
Made one mod to flatten company level departments. Looks great, thank you so much!!!

PR #3818, opened 2022-06-07
Update link to our jobs page.

mplacona:
The previous link pointed to https://about.sourcegraph.com/careers which is now obsolete. So we're now using https://about.sourcegraph.com/jobs instead.


InesRoitman (approved):
LGTM thanks :)

Issue #3838, opened 2022-06-08
Add linter to avoid insecure links to content


elzannewentzel:
add a linter to the handbook to ensure people don’t add insecure links (http) to the content.

![Image](https://user-images.githubusercontent.com/88658267/172621591-db5ddc46-e0e5-4d5f-b398-57c4ba80b1e6.png)

![Image](https://user-images.githubusercontent.com/88658267/172621628-5e4041c7-a3c1-4cda-877e-d09427e293ed.png)

PR #3839, opened 2022-06-08
Fix http image url

marekweb:
Fix an image url that was using HTTP instead of HTTPS


bretthayes (approved):
LGTM! We have #3838 to follow up with this to prevent http links from sneaking in.

PR #3914, opened 2022-06-14
Update QBR template

malomarrec:
From what I understand, this is the new template.
Squash and merge upon approval.


amenne:
This is the template when Value Engineering is performing Value Realization during the QBR. More often than not this does not occur so the other template is the one we use.

malomarrec:
@amenne better?

RoudyBob (approved):

malomarrec:
Posted [here](https://sourcegraph.slack.com/archives/CU93UDUBV/p1655382121183889), merging.

PR #3955, opened 2022-06-16
update: managed instance upgrade list

kalanchan:


michaellzc:
I think trivial handbook changes we can just merge on our own without approval. high agency ;)

michaellzc (approved):

PR #3983, opened 2022-06-20
incidents: update post-mortem flow to ensure GitHub issues are linked

bobheadxi:
This is in response to [a recent request that all follow-ups in incident.io are checked to be resolved](https://sourcegraph.slack.com/archives/C02KZ941RNX/p1655726416453319) - the process does not yet make it clear that follow-ups should be created and should be tied to GitHub issues that can be synchronized by incident.io.


jeanduplessis (approved):
LGTM 🚀

PR #3995, opened 2022-06-22
Small formatting changes to incident process

marekweb:
Make the process steps consistent with the writing style/format of other incident process steps on the page.


bobheadxi (approved):
Nice! Ths is much more readable

PR #4000, opened 2022-06-22
Update team for repo deletion

gabtorre:


airamare01 (approved):
Thanks for catching this @gabtorre!

PR #4026, opened 2022-06-23
update CSP to allow Github resources (and team map)

cconcannon:


cconcannon:
Here's a screenshot to illustrate why the update to `netlify.toml` is needed - the content security policy doesn't currently allow fetching of the embedded team locations map from `embed.github.com`

Screen Shot 2022-06-23 at 1 30 10 PM

cconcannon:
fixed
Screen Shot 2022-06-23 at 2 05 48 PM

PR #3958, opened 2022-06-16
add list of external calls

Joelkw:
Based on this thread – https://sourcegraph.slack.com/archives/C02KZ941RNX/p1654758176920729?thread_ts=1654621378.288329&cid=C02KZ941RNX

PR #4034, opened 2022-06-24
Put customer interview notes in a single folder

malomarrec:
Today, long customer interview transcripts don't fit in productboard cards. So longer interview stay in a personal drive and there's no way to find them.
Productboard is the SoT for feedback, but it benefits everybody to put long google docs in the same folder so that they don't get lost.
[Thread](https://sourcegraph.slack.com/archives/C0W2E592M/p1656104783143769?thread_ts=1654708469.429829&cid=C0W2E592M)

### DIFF
--- a/content/departments/engineering/product/process/user_stakeholder_feedback.md
+++ b/content/departments/engineering/product/process/user_stakeholder_feedback.md
@@ -110,6 +110,10 @@ See [How to reference customer names in public tickets](prioritizing.md#how-to-r
 
 We aggregate all of this feedback in [productboard](https://sourcegraph.productboard.com/). This lets us look at all of the sources of feedback to better understand which features will have the biggest impact on the product. It helps with prioritization and making sure we're building the next most important thing. It also helps us maintain a long-term source of truth for feedback: some of our sources of feedback are subject to removal over time due to retention policies, so adding specific insights to productboard captures it for the future.
 
+In case you have longer google docs (eg. customer discovery notes) that don't fit in productboard cards:
+- put the doc in the [customer feedback notes](https://drive.google.com/drive/folders/1suKSsUTTJlEyRaiunNtzJUMwrP_ZPWYG) folder, using the following naming convention `Customer Year - <notes>` eg. `ACME Corp 2022 - code security notes`.
+- create snippets for key insights into Productboard
+
 ### Adding feedback to productboard
 
 All of the above sources of feedback (with the exception of HubSpot forms) require the team help forward that data into productboard using an available integration:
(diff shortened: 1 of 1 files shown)


ryankscott:
@malomarrec - I wonder if this should be added to [`content/departments/engineering/product/design/research/index.md`](https://github.com/sourcegraph/handbook/blob/00e723b467f2f830035c33eb876a1a4ab9970bb8/content/departments/engineering/product/design/research/index.md) instead? 
This is where @Joelkw's change [on scheduling interviews will be added](https://github.com/sourcegraph/handbook/pull/3958).

malomarrec:
I don't disagree.
Should research be promoted as a top level article? I don't think research
should only be a design team thing TBH

malomarrec:
Wondering if we should not put it here:
https://github.com/sourcegraph/handbook/tree/00e723b467f2f830035c33eb876a1a4ab9970bb8/content/departments/engineering/product/process/user_research

ryankscott:
Yeah agreed it shouldn't be just a design thing. Great call on putting it there, that makes the most sense. 

We can work with the design / research team to move all of their relevant info under that page.   FYI @meglynst

content/departments/engineering/product/process/user_stakeholder_feedback.md
```diff
@@ -110,6 +110,11 @@ See [How to reference customer names in public tickets](prioritizing.md#how-to-r
 
 We aggregate all of this feedback in [productboard](https://sourcegraph.productboard.com/). This lets us look at all of the sources of feedback to better understand which features will have the biggest impact on the product. It helps with prioritization and making sure we're building the next most important thing. It also helps us maintain a long-term source of truth for feedback: some of our sources of feedback are subject to removal over time due to retention policies, so adding specific insights to productboard captures it for the future.
 
+In case you have longer google docs (eg. customer discovery notes) that don't fit in productboard cards:
+
+- put the doc in the [customer feedback notes](https://drive.google.com/drive/folders/1suKSsUTTJlEyRaiunNtzJUMwrP_ZPWYG) folder, using the following naming convention `Customer Year - <notes>` eg. `ACME Corp 2022 - code security notes`.
```
Joelkw:
```suggestion
- put the doc (or a [shortcut](https://support.google.com/drive/answer/9700156?)) in the [customer feedback notes](https://drive.google.com/drive/folders/1suKSsUTTJlEyRaiunNtzJUMwrP_ZPWYG) folder, using the following naming convention `Customer Year - <notes>` eg. `ACME Corp 2022 - code security notes`.
```

Shortcut because there may be things that go in other folders or are already in other folders and people may look for them there (e.g. code insights teammates know to check our google drive code insights folder for these notes, and while I could shortcut to the code insights folder if we prefer that, to me it feels like the "horizontal" folder is the one that should have shortcuts, which is this).

Joelkw (approved):
Thank you!

content/departments/engineering/product/process/user_stakeholder_feedback.md
ryankscott: Nit: It'd be nice to standardise on `YYYYMMDD - Customer - Subject` (e.g. 20222806 - Acme - Code security). Only for sorting purposes.

ryankscott (approved):
LGTM - We just need to work out the best place for this information. Even here is fine until we rework the information hierarchy.

content/departments/engineering/product/process/user_stakeholder_feedback.md
malomarrec: Great idea

meglynst:
@ryankscott @Joelkw @malomarrec I'm starting work to make research processes more efficient, which includes rework of the research part of the handbook as well as figuring out a more dedicated place for research artifacts and insights like this to live.

PR #4091, opened 2022-06-29
Update product_teams.yml

jttaylor210:
added brief descriptions for the different product teams

I think it would be a good idea to include a brief description on this page to allow visitors to know which team does what to prevent them from needing to click into the actual strategy page. 

This would reduce a click for someone looking to quickly understand one product over the other and guide them to look into the product they're most interested in


jttaylor210 (commented):
Pushing

PR #4102, opened 2022-06-30
...

rafax:
[Issue](https://github.com/sourcegraph/customer/issues/1108)


sanderginn:
Just a heads up that it's recommended not to include the customer name in commits on the handbook

PR #4123, opened 2022-07-01
Flatten strat pages, remove org strat pages

malomarrec:
This flattens the strat part of the handbook.
Org pages in engineering/product are unchanged.


Joelkw (approved):
TY!

PR #4141, opened 2022-07-05
Proposal: remove IC5 profiles

mrnugget:
This proposal is the result of yesterday's and last week's IC5 sync in which we agreed that the IC5 profiles cause us more confusion than anything else. Confusion comes from us asking us the following questions constantly:

- are the profiles a set of expectations that we need to live up to? or a description of the work we're already doing?
- what if we want to do something different, something that doesn't match the profile anymore? does the profile need to be rewritten? do we lose the title of IC5?
- in some cases the profiles were reverse-engineering descriptions of what individuals were already doing - does that make the profiles unfit to lead and serve as a north star?
- in other cases the profiles are really abstract and neither serve as a north star nor as a description that other ICs can look to as guidance for what IC5-level work looks like
- ...

We do not yet know how to further shape the IC5 role, whether we'll find something to replace the IC5 profiles or not, whether we'll adopt industry-standard terminology (staff eng & archetypes such as: tech lead, architect, ...), or do something else.

This is meant as a first step _back_ so we can decide which two steps forward we want to take.

It's a *proposal* too: this PR should serve as a discussion and if someone disagrees with this proposal and we end up having a discussion around IC5 profiles, it served its purpose.

If we (the IC5s) all agree on the removal, we'll ask Quinn for his approval.


keegancsmith (approved):

rafax (approved):

sqs (approved):

efritz (approved):

rvantonder (approved):
> This is meant as a first step back so we can decide which two steps forward we want to take.

To prompt some discussion here, I think it's valuable to revisit why these profiles exist at all. My understanding is partial, but I do remember some details when this change was in flight:

- progress on defining IC5 (or "level above/aside from IC4") was stagnant for many many months. This started to cause real issues and discrepancies when...

- we clearly had a need to hire and/or promote at the "level above/aside from IC4"

Again to my understanding, profiles were a way to communicate and convince key stakeholders that the role should exist to serve specific business needs. That move had the desirable effect of unblocking some of the underlying issues above re: (1) lack of progress to define "higher/different level" (2) need for "higher/different level" to exist as a hiring or promotion band

I'm approving from a personal "I believe profiles served its purpose to enable progress here, and we do not need them to underscore business need by specificity".

I'd like @lguychard to give his two cents, since he's more intimate with the original details / reasoning of profiles. My main question to Loic is: do we stand to lose anything significant (organizationally, or stunting career framework progression, hiring needs, or evaluating bands, etc.) by removing profiles?

efritz:
> Again to my understanding, profiles were a way to communicate and convince key stakeholders that the role should exist to serve specific business needs.

@rvantonder Are any of the key stakeholders for which the business profiles mattered at the company anymore? (This isn't snark, it's a genuine question that came up during our IC5 meetings the past two weeks.)

rvantonder:
>  Are any of the key stakeholders for which the business profiles mattered at the company anymore?

I do not know. Which is why I'm calling on @lguychard (as a key driver) to cover this case statement :-). Because what I do know is that a lot of effort went into it to overcome whatever heavy barriers we were facing at the time that dragged out this problem (and at some level, Quinn has oversight of such barriers and decision making).

lguychard:
I documented the [historical context](https://docs.google.com/document/d/1m1ssgp6Fu3zrVoW9SQ6Re0WZEcXyVIfKzsvHJmpVRRw/edit) that led to creating these profiles, and Quinn is aware of said context.

> do we stand to lose anything significant (organizationally, or stunting career framework progression, hiring needs, or evaluating bands, etc.) by removing profiles

Given that a) our current eng decision-makers are aware of the historical context and b) we have a simpler structure now, where IC5s themselves have the agency to effect change, I have no concerns about moving forward with this change.

mrnugget:
I'm going to go ahead and merge this. We have two outstanding reviews (Michal and Olaf are out), but they were in the IC5 sync calls in which it was agreed to drop the profiles. We can always revert and discuss.

mrnugget:
(@olafurpg & @vrto: I only removed you as reviewers in the hopes that this would make auto-merge work, since I can't manually merge here, but seems like I only had to rebase 🤦 )

olafurpg (commented):
Belated LGTM 👍

PR #4151, opened 2022-07-05
NPS widget collects use cases, has two channels

malomarrec:
* from Sourcegraph 3.41, the NPS widget collects use case data
* NPS scores are posted in #feedback-nps, longer submissions in #feedback


umpox (approved):
Thanks!

PR #4152, opened 2022-07-05
code security tiger team wrapup

malomarrec:
* remove tiger team pages
* add key learnings to use case page


sqs (approved):
There are 2 yaml files that need to be edited as well: https://sourcegraph.com/search?q=context:global+repo:sourcegraph/handbook+code+security+tiger&patternType=literal

malomarrec:
Dammit I forgot to save those. Good catch, ty!

PR #4178, opened 2022-07-06
Update colors to reflect DLS.md

traceyljohnson:
I updated colors to align with the DLS. 
Including:
- updated color names
- added a few new colors that were missing

Still need to add grayscale, but this is a start!

### DIFF
--- a/content/departments/engineering/design/brand_guidelines/color.md
+++ b/content/departments/engineering/design/brand_guidelines/color.md
@@ -82,11 +82,11 @@ Our colors express our commitment to being a welcoming, vibrant, and accessible 
 
 <div class="color-palette interactive">
   <div class="color" style="--color: var(--sg-sky-blue)">
-    <h5>Sourcegraph Sky Blue</h5>
+    <h5>Sourcegraph Blue</h5>
     <code>#00cbec</code>
   </div>
   <div class="color" style="--color: var(--sg-vivid-violet); color: white">
-    <h5>Sourcegraph Vivid Violet</h5>
+    <h5>Sourcegraph Violet</h5>
     <code>#a112ff</code>
   </div>
   <div class="color" style="--color: var(--sg-vermillion); color: white">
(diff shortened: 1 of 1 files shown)


content/departments/engineering/design/brand_guidelines/color.md
```diff
@@ -107,130 +107,200 @@ Primary logo colors have been marked. The secondary colors, along with shades an
...
+      <small>Grayscale 500</small>
+    </div>
+    <div class="color color-sm" style="--color: #313131; color: white">
+      <code>#313131</code>
+      <small>Grayscale 600</small>
+    </div>
+    <div class="color color-sm" style="--color: #000000; color: white">
+      <code>#000000</code>
+      <small>Black</small>
     </div>
```
st0nebreaker, line 304: Looks beautiful, thank you! Do you want the grayscale blocks to wrap like the colored ones are? If so, I'm happy to do that real quick @traceyljohnson 
image

content/departments/engineering/design/brand_guidelines/color.md
```diff
@@ -82,11 +82,11 @@ Our colors express our commitment to being a welcoming, vibrant, and accessible
 
 <div class="color-palette interactive">
   <div class="color" style="--color: var(--sg-sky-blue)">
-    <h5>Sourcegraph Sky Blue</h5>
+    <h5>Sourcegraph Blue</h5>
```
felixfbecker, line 85: If we rename a color, we also need to rename the CSS variable (see right above this line). It's defined in [layout.scss](https://sourcegraph.com/github.com/sourcegraph/handbook/-/blob/src/styles/layout.scss?L49). These variables are used within the handbook itself, otherwise it becomes very confusing to use them.

content/departments/engineering/design/brand_guidelines/color.md
```diff
@@ -240,58 +310,58 @@ Primary logo colors have been marked. The secondary colors, along with shades an
 Use supporting colors for added tonality and awareness within website UI.
 
 <div class="color-palette interactive" style="max-width: 54rem">
-  <div class="color color-lg" style="--color: var(--sg-light-gray)">
-    <h5>Sourcegraph Light Gray</h5>
+  <div class="color color-lg" style="--color: #f9fafb">
+    <h5>SG Light gray</h5>
```
felixfbecker, line 314: Our terminology/copy guidelines say to not abbreviate Sourcegraph to SG: https://handbook.sourcegraph.com/company-info-and-process/communication/content_guidelines/terminology_guidelines/#yes-1
(this doesn't apply to code, but I think it should apply to handbook content)

felixfbecker (commented):
I just saw these changes now, so sorry for the late review, but saw some things necessary to point out. I wrote the initial version of these pages so happy to help.

PR #4211, opened 2022-07-07
update-managed-instance-AER-handbook

DaedalusG:
Bring handbook up to date with current managed instance practices


jasonhawkharris (approved):
LGTM!

PR #4213, opened 2022-07-07
Add instructions for reissuing an expiring license key.

alexfogg:
Add documentation for re-issuing a license key that is about to / has expired.

### DIFF
--- a/content/departments/ce-support/ce/process/license_keys.md
+++ b/content/departments/ce-support/ce/process/license_keys.md
@@ -10,6 +10,14 @@ For a new trial, create a _new_ subscription. You will only use this for the tri
 
 When the customer converts, create a _new_ subscription; do not reuse the trial subscription. (Follow the instructions below from start to finish.) This is to ensure data accuracy for the Sales Ops team.
 
+## How to reissue a license key for an expiring license
+
+In most circumstances, license keys match the renewal dates on an account contract and are generated well ahead of time. Sometimes, license keys deviate from this format due to product trials and other special circumstances. If a license key is imminently expiring or has expired, follow these steps:
+
+1. If possible, check with the account's Customer Engineer & Account Executive, as well as the account running notes. This will help you understand any special context surrounding the current license key tags and expiration dates. The account team may opt to extend a product trial key for a specific amount of time.
+2. Grab the contract end date by going to the Account page in Salesforce to look up the `"Active Contract Period End"`.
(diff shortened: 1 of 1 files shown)


content/departments/ce-support/ce/process/license_keys.md
```diff
@@ -10,6 +10,14 @@ For a new trial, create a _new_ subscription. You will only use this for the tri
 
 When the customer converts, create a _new_ subscription; do not reuse the trial subscription. (Follow the instructions below from start to finish.) This is to ensure data accuracy for the Sales Ops team.
 
+## How to reissue a license key for an expiring license
+
+In most circumstances, license keys match the renewal dates on an account contract and are generated well ahead of time. Sometimes, license keys deviate from this format due to product trials and other special circumstances. If a license key is imminently expiring or has expired, follow these steps:
+
+1. If possible, check with the account's Customer Engineer & Account Executive, as well as the account running notes. This will help you understand any special context surrounding the current license key tags and expiration dates. The account team may opt to extend a product trial key for a specific amount of time.
```
alexfogg, line 17: Obviously best future case is the concept of a product trial is totally separate from the core Sourcegraph license, but let's work with what we have for now.

Do we want to add any additional instructions for the cases where it _is_ a code insights/batch changes trial? We ideally wouldn't want to accidentally add it to their annual license; but we don't have a standardized way of flagging this right now. (Maybe with Vitally we can? Or make everyone add this to their running notes in the meantime?)

amenne: We will be moving to product specific subscriptions in the near future and away from this account level single license key structure. It does mean that we'd have licenses per product subscription so I think the guidance is more whatever license is expired... does that make sense?

content/departments/ce-support/ce/process/license_keys.md
```diff
@@ -10,6 +10,14 @@ For a new trial, create a _new_ subscription. You will only use this for the tri
...
+2. Grab the contract end date by going to the Account page in Salesforce to look up the `"Active Contract Period End"`.
+3. Follow the process below for generating a renewal license key. Unless otherwise noted during Step 1, use the same exact tags as the existing key and use the `"Active Contract Period End"` as the expiration date.
```
amenne: "use the same exact tags (make sure to comma separate each tag)"

amenne (approved):
One comment and one recommended change otherwise lgtm

PR #4227, opened 2022-07-08
Color palette min-width

st0nebreaker:
Sets a minimum width & removes gap spacing for Secondary colors. Allows names & blocks better UI.

**Note**: The secondary color columns stay on the same row until viewport width is below ~1188px.

**Staging link**: [here](https://deploy-preview-4227--sourcegraph-handbook.netlify.app/departments/engineering/design/brand_guidelines/color/).

image


traceyljohnson (approved):
Gorgeous!!

content/departments/engineering/design/brand_guidelines/color.md
```diff
@@ -37,6 +37,7 @@ Our colors express our commitment to being a welcoming, vibrant, and accessible
   }
   .color-palette .color-sm {
     width: 6rem;
+    min-width: 104px;
```
felixfbecker, line 40: The line right above this assigns an explicit width of 6rem (96px), and this line then forces a min-width of 104px (6.5rem), meaning the line above would never be in effect. It would be better to update the `width` to `6.5rem` directly (or even `7.25rem`, to fill out he page width better)

content/departments/engineering/design/brand_guidelines/color.md
```diff
@@ -103,8 +104,8 @@ Use supporting colors for graphic elements, illustrations, callouts, website UI
 
 Primary logo colors have been marked. The secondary colors, along with shades and tints, were chosen to support and compliment these three core colors.
 
-<div class="color-palette interactive">
-  <div class="color-palette-column">
+<div class="color-palette interactive" style="gap: 0rem">
```
felixfbecker, line 107: The `gap` comes from the CSS `<style>` right above on the top of the page – instead of putting a style override on every individual block, it would be better to just remove the `gap` declaration in the CSS `<style>` block so it's consistent everywhere and doesn't require duplication. Although I would suggest at least making the gap `0.125rem` for readability.

felixfbecker (commented):
I just saw these changes now, so sorry for the late review, but needed to point out some things. I wrote the initial version of these pages so happy to help.

PR #4285, opened 2022-07-13
updated Which MFA options are available?

nickyvm:
I added some language about TOTP and MFA options


cconcannon (approved):
The MFA instructions option descriptions and instructions are clear.

PR #4309, opened 2022-07-18
Refactor for security pages

evict:


ferozsalam:
This looks good! A much more focused landing page than we had before. The only change I was wondering about is whether we might want to also move the ambassador section into a separate document, perhaps the 'What we do' page?

mohammadualam (approved):

PR #4310, opened 2022-07-18
add mention of 4.0

benvenker:


sqs (approved):
- approving to make it clear it's fine to mention 4.0 in public (although for anyone else reading, we should do so in internal transparent docs like the handbook for internal purposes, not start to boast about it or pre-announce it publicly)
- but we have the August release between now and 4.0 (ignoring the July and September releases), and it would be great for some of the obvious, incremental things to ship in that.

benvenker:
@sqs sounds good. Things will begin shipping in this upcoming release behind the toggle, so we've got incremental releases covered.

PR #4412, opened 2022-07-23
Remove server-side batch changes flag

malomarrec:


content/departments/cloud/technical-docs/v1.1/mi1-1_enable_executors_process.md
```diff
@@ -114,6 +114,10 @@ Link Monitoring email notification channels we created from the terraform module
 
 ![create-budget-02](https://storage.googleapis.com/sourcegraph-assets/create-executor-budgets-02.png)
 
+### Remove the server-side batch changes feature flag
```
filiphaftek: I would propose to put it before [Wrapping up](https://github.com/sourcegraph/handbook/blob/bab3b09f83ae43488d16e313fc3ad7b13ad4dc6c/content/departments/cloud/technical-docs/v1.1/mi1-1_enable_executors_process.md#wrapping-up) section, b/c now is inside billing, and this is not billing related, right?

filiphaftek (approved):
@malomarrec thx for that 👍 I would just propose to move this section before `Wrapping up` section, to make sure we do not miss it.

PR #4435, opened 2022-07-26
Added target companies list

Kelseynagel:
I linked and added our target companies list for recruiters


Kelseynagel:
added target companies list!

PR #4447, opened 2022-07-27
Add design project templates

rrhyne:
Adds design project templates to be used in creating design issues and a link to them via the design process.


.idea/modules.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
danielmarquespt: Should these .idea files be part of the Pull Request?

danielmarquespt (approved):
Thanks for adding these to the handbook. The only comment I have is about the . files that are part of this PR

danielmarquespt:
Also, should we add these here? https://github.com/sourcegraph/sourcegraph/tree/main/.github/ISSUE_TEMPLATE

It would be helpful, but at the same time it would create a lot of options that are only used by designers here https://github.com/sourcegraph/sourcegraph/issues/new/choose

.idea/modules.xml
rrhyne: Thanks, they should not. I removed and added .idea to the .gitignore

rrhyne:
> Also, should we add these here? https://github.com/sourcegraph/sourcegraph/tree/main/.github/ISSUE_TEMPLATE
> 
> It would be helpful, but at the same time it would create a lot of options that are only used by designers here https://github.com/sourcegraph/sourcegraph/issues/new/choose

I chose not to because many developers feel like we have too many issue templates.

content/departments/engineering/design/templates/extra-large-template.md
```diff
@@ -0,0 +1,68 @@
...
+  - These efforts require close collaboration with the Product Manager to scope and define the user problems being solved.
+
+✂️ -------------------------
+
+```markdown:
+## Design process
+
+  - Estimate: More than a month
+  - Weighting factors:
+    - Note: a project of this size requires resources that increase its risk profile.
```
jjinnii, line 19: What does it mean to provide resources that increase a risk profile?

rrhyne: More time and more people increase risk.

PR #4503, opened 2022-08-01
chore: adjust dates per handbook guidance

unknwon:
We should be using the format `YYYY-MM-DD` per [our handbook](https://handbook.sourcegraph.com/company-info-and-process/communication/content_guidelines/style_and_mechanics/#dates), the current format also confuses our tooltip:

CleanShot 2022-08-01 at 14 17 26@2x


akalia25 (approved):
LGTM

PR #4527, opened 2022-08-02
Changing name of CTE landing page

emchap:
Renaming the CTE landing page in an attempt to resolve handbook 404 issue, per https://sourcegraph.slack.com/archives/C015P61EXV1/p1659046185355739


emchap:
@felixfbecker Check is failing in a way that suggest that the rename isn't triggering an auto-redirect. Can you help here?

emchap:
I can edit the offending link, but want to make sure I understand how redirects are meant to work in the future.

felixfbecker:
The internal links always need to point to the new/updated locations. This is so that even when viewing the markdown files on GitHub or in pull requests (where there are no redirects) links work correctly. Any links from outside the handbook though will be redirected properly.

emchap:
Gotcha. In that case, consider this a feature request for that—when I asked about redirects, I was interested in internal redirects, since internal links are much more common than external ones, at least for any content for my team.

PR #4537, opened 2022-08-03
Update `mg` installation guide

michaellzc:


content/departments/cloud/technical-docs/operations.md
```diff
@@ -27,6 +27,10 @@ Below we will install `mg` CLI. `mg` simlifies operation on MI and releases the
 ```sh
 mkdir -p ~/.bin
 export GOBIN=$HOME/.bin
```
daxmc99, line 29: We shouldn't be making our own bin folder... 😕 
https://go.dev/doc/gopath_code#GOPATH
We should install to `~/go/bin` (like other Go binaries) and then ask that users ensure that their GOBIN is first on their path

content/departments/cloud/technical-docs/operations.md
```diff
@@ -27,7 +27,12 @@ Below we will install `mg` CLI. `mg` simlifies operation on MI and releases the
...
+# ZSH users: echo "export \$PATH=\$HOME/.bin:\$PATH" >> ~/.zshrc
+# you need ensure our `mg` binrary has the highest priority in $PATH
```
daxmc99:
```suggestion
# you need ensure our `mg` binary has the highest priority in $PATH
```

leonore:
```suggestion
# you need ensure our `mg` binary has the highest priority in $PATH
```

leonore (approved):
typo fix but lgtm!

content/departments/cloud/technical-docs/operations.md
michaellzc: I am not sure why asdf doesn't configure the $GOBIN env var for me, and I always like having my own bin folder for random binaries.

daxmc99: I would like `mg` to be proper Go binary at some point 😓  
@danieldides set this path, is there a reason we used this instead of `$(go env GOPATH)/bin`?

danieldides: I think it used to be recommended to explicitly set a `$GOBIN`, `$GOPATH`, etc in your shell profile. I totally forgot that was something that I [had configured](https://github.com/danieldides/dotfiles/blob/master/zsh/.zshrc-common#L10) rather than something that ships by default. `$(go env GOBIN)` is probably more portable than relying on the existence of `$GOBIN`. This just needs to be updated in the `Makefile`.

Personally I like having it alongside other Go binaries but it doesn't really have a huge impact on end-user experience. It looks like `sg` [installs to](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/dev/sg/sg_install.go?L75) `/usr/local/bin` which may be a good option as well as most people should have `sg` installed and properly configured their paths to execute it.

At some point we're probably going to have to separate `mg` from the customer data, at which point we can properly relocate it to `cmd/mg/` so that we can `go install` it directly.

daxmc99 (approved):
Don't let my nits block 😃

PR #4601, opened 2022-08-10
Things we should consider for everything we ship (new and existing features)

sqs:
These are the things we should consider for everything we ship (new and existing features).

I've had this list in my mind for a couple weeks and wanted to write it down so that we can (as an eng dept) align on these considerations/principles.

Eventually this could become something like a checklist, but for now (until we've worked with these a bit more), let's keep it as high-level principles.


malomarrec:
@sqs I suggest breaking it down into smaller logical sublists for readability https://github.com/sourcegraph/handbook/pull/4603

malomarrec:
I think it'd be useful to have some form of guideline for experimental features, beta features, and GA.

For example, requiring that all new features support 100,00 scale or monorepo will make development much slower. So what are the requirements for shipping a feature, and how does that move with experimental/Beta/GA?

Where do we document that some features only work up to `n` repos, or in certain circumstances, and when is that acceptable? In the docs? Or in a [matrix](https://handbook.sourcegraph.com/departments/engineering/product/tools/)

malomarrec:
I think it'd be useful to have some form of guideline for what level we expect of features depending on their stage (experimental features, beta features, and GA), and how we communicate that (also see [docs](https://docs.sourcegraph.com/admin/beta_and_experimental_features)) 

For example, requiring that all new features support 100,00 scale or monorepo will make development much slower. So what are the requirements for shipping a feature, and how does that move with experimental/Beta/GA?

Where do we document that some features only work up to `n` repos, or in certain circumstances, and when is that acceptable? In the docs? Or in a [matrix](https://handbook.sourcegraph.com/departments/engineering/product/tools/)?

That may be a bit too operational for this page, but at our scale I think we need to think about how we operationalize this and communicate this internally (eg. to CEs and CSEs) and externally.

One way to make this work in practice is to say "we only ship to the smallest amount of customers first". So when a new feature comes out, we could in most cases first ship it to only cloud instances, feature flagged off by default. Then turn the feature flag on by default. Then ship it in self-hosted, etc.
I think it's better to ship down scoped features to a small amount of customers fast, then grow the target audience as the feature matures.

content/departments/engineering/index.md
```diff
@@ -51,6 +51,34 @@ The following functional teams work across the above dev teams:
 - [Working with issues](working-with-issues.md)
 - [Our software development lifecycle (SDLC)](sdlc.md)
 
+## What's in a feature?
+
+For every feature we ship, consider:
```
scalabilitysolved, line 67: Is this an ordered list?

Joelkw, line 67: what does it mean to "consider" – is this a requirement?

 If it's not a requirement, how should people use this? 

If it is a requirement, can we provide staged tiers (specifically around scalability – neither batch changes nor code insights works well for 100ks+ repos right now, but obviously they still provide tons of value). I'm happy to own defining tiered scale support here on this page or somewhere else if that'd be valuable.

content/departments/engineering/index.md
```diff
@@ -51,6 +51,34 @@ The following functional teams work across the above dev teams:
...
+- **Iteration feedback:** Iterate with our existing customers and gather user feedback as early as possible.
+- **Usage/value analytics:** Show how the feature is being used to users and admins, communicate the value to the customer's business, and identify opportunities for more value.
+- **Accessibility:** Make it accessible to a broad set of users (including keyboard accessibility).
+- **Marketing:** Update all marketing content to reflect the new feature (such as our about site, readmes, extension descriptions, slides, etc.).
+- **Observability:** Help admins debug, manage, and scale the functionality.
+- **Telemetry:** Gather the (minimum) data needed for us to understand how the feature is being used and improve it.
+- **Backward-compatibility:** Make it smooth to upgrade from past versions and adapt usage to the newest version.
+- **Forward-compatibility:** Anticipate the feature's future evolution (where possible, and without over-building today).
+- **Pricing/packaging:** Decide which plans include this feature and whether its code will be open-source or not.
```
jjeffwarner, line 91: I think Reliability touches on this but would like to see a specific call out to *Testing* - How we will test a feature so we understand how it behaves.

content/departments/engineering/index.md
```diff
@@ -51,6 +51,34 @@ The following functional teams work across the above dev teams:
...
+
+- **Security:** Make it secure, because our customers trust us with very sensitive code and other data, and we need to continually earn and retain that trust.
+- **Reliability:** Make it stable and robust, because devs don't trust flaky or incorrect results.
+- **Speed:** Make it fast, because devs won't use something 10+ times daily if it's slow.
+- **Scalability:** Make it work at large scale, such as big monorepos or 100,000s of repositories.
+- **Documentation:** Add and update documentation for the new feature.
+- **Announcement:** Draft the eventual announcement of the feature early (in blog post form, along with a changelog entry and major docs changes), and update it as you iterate.
+- **Iteration feedback:** Iterate with our existing customers and gather user feedback as early as possible.
+- **Usage/value analytics:** Show how the feature is being used to users and admins, communicate the value to the customer's business, and identify opportunities for more value.
+- **Accessibility:** Make it accessible to a broad set of users (including keyboard accessibility).
```
felixfbecker, line 76: When talking about accessibility, I worry that "a broad set of users" is (in many minds) satisfied even by just targeting the "common" user with no impairments (i.e., basically _not_ caring about accessibility). If we want to state this as a goal, I think we need to make that stronger
```suggestion
- **Accessibility:** Make it accessible to all users (including keyboard accessibility).
```

content/departments/engineering/index.md
```diff
@@ -51,6 +51,34 @@ The following functional teams work across the above dev teams:
 - [Working with issues](working-with-issues.md)
 - [Our software development lifecycle (SDLC)](sdlc.md)
 
+## What's in a feature?
+
+For every feature we ship, consider:
+
+- **Security:** Make it secure, because our customers trust us with very sensitive code and other data, and we need to continually earn and retain that trust.
+- **Reliability:** Make it stable and robust, because devs don't trust flaky or incorrect results.
+- **Speed:** Make it fast, because devs won't use something 10+ times daily if it's slow.
```
felixfbecker, line 73: I really like how all statements up to this one end with a "[...], because [...]". Makes it easy for everyone to empathize with why it's important for customers. Maybe we could do it for more points?

unknwon (approved):

content/departments/engineering/index.md
sqs: No (except that putting security first was intentional).

content/departments/engineering/index.md
sqs: Can you send a followup PR with what you'd propose here? I see testing as "how to achieve these things" rather than a goal in itself, but it still might be worth calling out specifically here.

PR #4638, opened 2022-08-12
Adding career frameworks definition

amath00:


felixfbecker:
Note that the page is still named `career-ladders`, I think that should probably be renamed

PR #4653, opened 2022-08-15
cloud: enable tracing on mi

danieldides:
Basic instructions to enable GCP tracing on managed instances. This will be on by default eventually but for now these instructions are sufficient.


content/departments/cloud/technical-docs/operations.md
```diff
@@ -432,6 +432,59 @@ resource.type="gce_instance"
...
+
+service:
+  extensions: [health_check, zpages]
+  pipelines:
+    traces:
+      receivers: [otlp]
+      exporters: [jaeger, googlecloud]
+```
+
+3. Update the corresponding `docker-compose.override.yaml` file to now include the `otel-collector` service and configure all existing services with the correct environment variables. Reference this [pull request enabling otel for `tpgi`](https://github.com/sourcegraph/deploy-sourcegraph-managed/pull/911/files) for an example of all that needs to be modified.
```
bobheadxi, line 471: Is this instance name sensitive information?

content/departments/cloud/technical-docs/operations.md
```diff
@@ -432,6 +432,59 @@ resource.type="gce_instance"
...
+4. Connect to the manage instance and add the following configuration to the `site-config`:
+
+```json
+
+"observability.tracing": {
+  "type": "opentelemetry",
+  "sampling": "selective",
+  "urlTemplate": "https://google.com?{{.TraceID}}"
```
bobheadxi: 🤯 this URL template just works?

michaellzc (approved):

content/departments/cloud/technical-docs/operations.md
danieldides: `tpgi` is an internally-used instance for validating accessibility, it's not a customer instance. That being said, I can remove the name since it's not necessary information and could be confusing for others.

content/departments/cloud/technical-docs/operations.md
danieldides: As far as I can tell, this field doesn't actually seem to be used with the `googleCloud` exporter. It must have the correct URL endpoints correctly configured. Traces with this configuration were correctly delivered.

bobheadxi: The URL template is for links _to_ the traces rendered in the UI, as well as occasionally for logs

danieldides: Ah. Then no, this URL will almost certainly not work. 🤣 I'll look into what the correct URL should be.

bobheadxi: Example: https://console.cloud.google.com/traces/list?project=sourcegraph-managed-sg&tid=166dd657862e1dda79009d5f0c78e523

If this works I think the template should be something like:

```
https://console.cloud.google.com/traces/list?referrer=search&tid={{.TraceID}}&project=sourcegraph-managed-TODO
```

PR #4678, opened 2022-08-18
Add design collaboration and review content, restructure to match

quinnkeast:
This content makes our design collaboration and review process more intentional. I anticipate that this draft will serve us immediately, and will evolve based on ongoing thinking from other design team members!


content/departments/engineering/design/design_collaboration_review/index.md
```diff
@@ -0,0 +1,115 @@
...
+
+Whether approaching [design collaboration](./#design-collaboration) or [review](./#design-review), we follow specific guidelines around [seeking and giving design feedback](./#seeking-and-giving-design-feedback).
+
+## Design collaboration
+
+Design collaboration is our chance to work together as designers on design efforts.
+
+Every design effort has a design owner. This person is usually the designer embedded on a product or feature team carrying out the effort.
+
+This design owner should identify a design collaborator near the beginning of the effort. This tends to make the most sense once the effort has gone through initial discovery and there’s a clearly-defined problem statement, and design exploration into the solution space has begun.
```
jjinnii, line 21: > This design owners _should_ identify a design collaborator....

Does this mean that we recommend having a design collaborator for all (most) of our design work? I feel like based on some of our recent design syncs, there is some ambiguity on when designers should be collaborating with another designer. 

Said another way, I feel like it's very clear to our team that we should continuously ask for design reviews, but it's undefined when design collaboration is _required_.

quinnkeast: I think *most* design efforts should have a design collaborator. 

- Some things are more intentional about this and may call for design collaborators to invest more time actually collaborating, with potentially more process between each other. This is particularly important when two people are working on different things at the same time with a lot of overlap/influence on each other.
- Other things are low-key, low-process, and fluid. Something that's bigger, but doesn't influence something else underway at the same time, still benefits from a design collaborator. Similarly, consider #40072—had I assigned Rob as a design collaborator directly, then I might've asked him while working on it for a quick gut check, and he might've replied with a quick "+1, better than current implementation."

Does this make sense? Is there anything I could change or add to make it more intuitive?

PR #4729, opened 2022-08-25
fix @leonore email to match slack

burmudar:
Build-tracker had some trouble to locate @leonore SlackID and would print `<@>`. After some digging I saw that the team.yml and what is in slack do not match 😄


leonore:
interesting. shame I can't use my alias for this

leonore (approved):

PR #4778, opened 2022-09-01
Clarify instance size in creation

michaellzc:


filiphaftek (approved):
Great!

content/departments/cloud/technical-docs/v1.1/mi1-1_creation_process.md
```diff
@@ -9,6 +9,18 @@ For basic operations like accessing an instance for these steps, see [managed in
 - `customer` - name of customer
 - `ce_email` - email of Customer Engineer from issue
 - `customer_email` - customer admin email (only one from provided in issue)
+- `instance_size` - the t-shirt size of the instance.
+
+  > review the eligibility of each instance size in the [Google Doc](https://docs.google.com/document/d/1WXfmyBsGlIoJvSdxaEhV0iYTD4ylm9ykLyJItDE53eE/edit#bookmark=id.hmlsv2azmjja)
```
RafLeszczynski, line 14: unless for some reason we do not want to have the instance sizes visible to the public I would recommend moving these definitions from google docs to the handbook. 

If we want to keep them internal only I recommend moving them to separate google doc

filiphaftek: @RafLeszczynski we have them defined in trials [here](https://github.com/sourcegraph/handbook/pull/4803/files#diff-533ec697163d9ed3416634f81491372a578f9114d9f0a18167cfb8c2f93b4490R43) - should we make them a link here and in trials?

PR #4799, opened 2022-09-05
Update Cloud FAQ with data migration answer

rafax:
Update Cloud FAQ to clearly state that Cloud doesn't support data migrations.


rafax:
Adding @sqs as Head-of-Eng/Head-of-Product - this is something we aligned on in the past, but it was never documented in Handbook.

RafLeszczynski (approved):

sqs (approved):

PR #4943, opened 2022-09-24
Update mechanics guide for cloud capitalization

iskyOS:
Adding HB guidance specific to using the word "Cloud" and when it should be capitalized.


iskyOS:
Also removed "Code Intelligence" as a product name while I was updating this page.

PR #4975, opened 2022-09-27
Cloud Trial: Fix indentations

unknwon:


unknwon:
I thought we need approves for merges but not anymore? Asked in https://sourcegraph.slack.com/archives/C07KZF47K/p1664246230982439

PR #4999, opened 2022-09-27
fix wording

malomarrec:


ngage1 (commented):
@malomarrec can you add a line after 27 that 'For now, All Trial requests will be processed by inbound SDR team

AND

Update our 'Slack Channel Creation' rules so that Inbound SDRs create a slack channel upon any approval, and will add the Account Team when:
- PQLd  if not a strat account or current customer
- Requests time with us

PR #5057, opened 2022-09-29
add noindex=true to NextSeo

cconcannon:


cconcannon:
This PR results in the following updates to `meta` tags which direct search engines to *not* index any page in the handbook.

```html
<meta name="robots" content="noindex,follow">
<meta name="googlebot" content="noindex,follow">
```

Screen Shot 2022-09-29 at 3 48 15 PM

cconcannon:
Closing because there was opposition to this idea, and nobody else supported it.

PR #5100, opened 2022-10-02
Add: Providing additional resources for troubleshooting

AlicjaSuska:
### Background
When working on a recent code insights project, I encountered a case: pointing a user to resources to help them fix an error/troubleshoot an issue. Our content guidelines do not cover it yet (or not fully). Let's align on a common language for those cases.

### More about the project:
A Code Insight (or a particular point) can time out. We need to inform the user about it directly on the insight card.
Many things could be causing the problem. It's outside our capability to point to an exact problem or fix it for the user.

We need to provide them with a link to docs that contain more details than we can express directly in the product so that they can troubleshoot the issue.
We're most probably need to express it in the tooltip with a link (see the attached screenshot).

Screenshot 2022-10-04 at 14 34 29

### The solution:
I've proposed an explicit 'Learn how to troubleshoot' message that expands beyond a simple 'learn more' but does not overpromise (like 'fix it' or 'get help' would).

### Similar cases
Linking to docs could also be used in the:
- alerts
- error messages (example: for the input fields) - especially for more complex cases when we know what the issue is, but it has multiple possible fixes

### Content guidelines
Currently, we cover error messages and links in our content guidelines. This case overlaps with those sections, but it is not covered explicitly. I've decided to add it as a separate section. Please, share if you have a better idea regarding the placement in the content guidelines.

### Requested feedback

1. Copy - is 'troubleshoot' a term we agree on? Should we consider any alternatives?
2. Placement in the content guidelines - separate section or one of the existing locations?


quinnkeast:
I think it's a good idea to have an explicit guideline for this case. My first reaction: "Learn more" is "`Verb` more", while "Learn how to troubleshoot" is "`Verb` how to `verb`." Might we simplify even further and use "Troubleshooting" alone? If so, we might be able to establish an additional pattern for these instances with a `→` or similar so that we don't have to use sentence case for this specific snippet of the overall string, and it gives us the flexibility to use more verbs than just learn more and troubleshoot. 

>Results may be incomplete. [Troubleshoot](#) →

>Limited to 100 results. [Learn more](#) → 

>Cannot connect. [Repair connection](#) →

felixfbecker:
> We're most probably need to express it in the tooltip with a link (see the attached screenshot).

Note that a [tooltip](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tooltip_role#:~:text=Because%20the%20tooltip%20itself%20never%20receives%20focus%20and%20is%20not%20in%20the%20tabbing%20order%2C%20a%20tooltip%20can%20not%20contain%20interactive%20elements%20like%20links%2C%20inputs%2C%20or%20buttons.)  (which is a defined term/pattern/component) cannot contain interactive elements like a link, because it pops up on focus or hover, and therefor cannot receive focus or hovers itself (as opposed to a popover).

> Because the tooltip itself never receives focus and is not in the tabbing order, a tooltip can not contain interactive elements like links, inputs, or buttons.

Do you think your example should use a popover, i.e. something that opens on click rather than hover and then receives focus, and would look visually distinct from our tooltip style?

AlicjaSuska:
@quinnkeast thank you for the suggestion. I've edited the examples to reflect the wording to be a simple verb ('troubleshoot').

@felixfbecker Thank you for pointing this out. I agree, let's make the popover (styled the same as toolip) appear on click on this icon. I've also noted this in the [designs](https://www.figma.com/file/aMO6sPoUHDUUcqa25wFnWl/Surface-timeout-limit-hit-to-end-users-in-the-UI-%2339969-Review?node-id=1842%3A4253).

PR #5156, opened 2022-10-05
Improve markdown formatting across a bunch of pages

felixfbecker:
A lot of pages in the handbook, especially recently added ones that were created from converting Google docs without polishing up the generated markdown, were very hard to read or edit. This fixes a bunch of them to follow best practices for markdown, like:

- Fix various accidentally missing line breaks that e.g. made different bullet points end up on the same line
- Fix various accidentally escaped formatting syntax `*` `_` that ended up in the output literally
- Don't link Slack channels manually, let the handbook do it automatically (better styling)
- Don't use underlines, which adds `<span style="text-decoration:underline;">` as HTML. Underlines are discouraged on the web due to looking like links. Use strong or em styling with `**`/`_`.
- Don't manually bold, underline or capitalize headings, as that makes headings inconsistent and the document structure hard to scan. Rely on the handbook to style headings appropiately to their level.
- Provide accessible, descriptive titles for links – don't name every link "here" or "click here"
- Use markdown within table cells, rather than having the whole table be HTML even for simple formatting within cells
- Use blockquote syntax/styling `>` for blockquotes
- Fix broken footnotes using real markdown footnotes
- Added some clear WIP markers where content was simply missing (e.g. sentences that ended mid-sentence)


jplahn (approved):
You beat me to a lot of these. Thanks for doing this! I did a super quick scan for anything that looked crazy.

PR #5251, opened 2022-10-13
Update data-analytics.md fix `capabilities` table to have references

akalia25:
Adds links to each bucket like it is on this page: https://handbook.sourcegraph.com/departments/data-analytics/


akalia25:
not needed since bizops pages are going to be deprecated

PR #5268, opened 2022-10-14
cloud/v1.1: update env setup

bobheadxi:
This is the new way to do env stuff as of https://github.com/sourcegraph/deploy-sourcegraph-managed/pull/1682


unknwon:
LOL I just had a PR that has the exact same diff as yours https://github.com/sourcegraph/handbook/pull/5270/files

But I forgot we don't have approval require for the handbook repository and enabled auto-merge 🤦‍♂️

PR #5324, opened 2022-10-19
MI: Update mi1-1_delete_process.md

unknwon:
I will never remember doing this first

### DIFF
--- a/content/departments/cloud/technical-docs/v1.1/mi1-1_delete_process.md
+++ b/content/departments/cloud/technical-docs/v1.1/mi1-1_delete_process.md
@@ -116,6 +116,8 @@ gcloud compute snapshots list --project=sourcegraph-managed-$CUSTOMER | grep "da
 ```sh
 # Under $CUSTOMER directory
 cd project
+
+terraform init # switch to local module
 terraform destroy -var-file=../terraform.tfvars
 ```
 


content/departments/cloud/technical-docs/v1.1/mi1-1_delete_process.md
```diff
@@ -116,6 +116,8 @@ gcloud compute snapshots list --project=sourcegraph-managed-$CUSTOMER | grep "da
 ```sh
 # Under $CUSTOMER directory
 cd project
+
+terraform init # switch to local module
```
filiphaftek, line 120: Just to clarify: init is not even about switching to local module, it's about initialise `.terraform` folder locally, so pulling all providers, libraries etc and connecting to GCS state bucket to check if all is working ;)

unknwon: LOL, that makes more sense, I literally copied the other one which has this comment. Do you think it is more clear without any comment? (I do)

filiphaftek: > LOL, that makes more sense, I literally copied the other one which has this comment. Do you think it is more clear without any comment? (I do)

No not, it's true what you said, b/c we use local module via `../modules./...` anyway - just wanted to mention it's not the main purpose of invoking this command ;)

filiphaftek (approved):

PR #5339, opened 2022-10-20
scaletesting: add auth command for the k8s cluster

leonore:
as the title says

also fix some emojis that don't get parsed in our handbook


burmudar (approved):
Thanks!

PR #5340, opened 2022-10-20
Adjust "wireframes" to "low fidelity design" in design issue templates

quinnkeast:
Found myself thinking "low fidelity design" is more open-ended and included wireframes as a technique, but not the only technique. More important is the idea that there's an exploration and collaboration phase ahead of high-fidelity specification.

Additionally, for `extra-small` issues, one might use high fidelity from the design system or a few low-fidelity specifications that developers can run with.


rrhyne (approved):
Makes sense

PR #5350, opened 2022-10-21
devx: update sections about devx operated instances

burmudar:
Added more detail about the instances of sourcegraph devx handles as well as add their deployment cadence


jhchabran (approved):
Nice! Thanks for taking care of this 🙏

PR #5355, opened 2022-10-21
fix doc link

taylorsperry:
Was linking to the wrong doc!


taylorsperry:
Thanks for catching, @felixfbecker !

PR #5443, opened 2022-10-31
cloud: add queries to show blocking queries

davejrt:
Add queries to list indexing and/or blocking services to unblock `migrator`.


burmudar:
Thanks for adding this!!! <3

PR #5453, opened 2022-10-31
mi1.1: add bulk upgrade tips

bobheadxi:


danieldides (approved):
This is great. I wonder if we shouldn't just render out these upgrade steps as part of our upgrade template and put the commands in the issue.

bobheadxi:
> This is great. I wonder if we shouldn't just render out these upgrade steps as part of our upgrade template and put the commands in the issue.

I can add that there too :)

PR #5517, opened 2022-11-07
Update the security team handbook

willdollman:


content/departments/security/index.md
```diff
@@ -4,15 +4,53 @@
...
+- [Report an incident](./security-incident-response.md#reporting)
+
+## Security Sub-Areas
+
+### Security Engineering
+
+### Security Operations
+
+### Security Assurance
+
```
evict, line 37: I think we cannot post it like this.

evict (changes requested):

PR #5570, opened 2022-11-15
Update db examples

gabtorre:


DaedalusG (approved):
Oh SQL command always gotta be caps 😆

airamare01 (approved):
Great catch!

PR #5661, opened 2022-11-25
Turn scaletesting helm chart version into variable

sanderginn:
No more cat and mouse with updating the version


burmudar:
Well this is NIFTY!

burmudar (approved):

PR #5673, opened 2022-11-28
Cleanup table markdown so it renders correctly

mike-r-mclaughlin:
The markdown in the table was wrong so it was rendering as raw text.


ferozsalam:
@mike-r-mclaughlin – there was actually an extraneous `j` in the table contents which was causing this issue – I have fixed it in #5675!

PR #5680, opened 2022-11-29
Instance Creation SLA is 1 working day for all instances

daxmc99:
We often create instances faster but this is a good upper bound for an
SLA


filiphaftek (approved):
👍

RafLeszczynski (approved):

PR #5763, opened 2022-12-08
banner: introduce Banners that can be added to subsets of pages

bobheadxi:
There is a risk of customers finding handbook content that does not accurately represent the current state of Sourcegraph Cloud offerings - to mitigate this while remaining open and transparent, we add prominent banners on every page in the `/cloud` path.

image

### DIFF
--- a/src/components/Banner.tsx
+++ b/src/components/Banner.tsx
@@ -1,0 +1,38 @@
+import React from 'react'
+
+/**
+ * Banner generates the appropriate banner(s) for the given slug paths. Banners are
+ * hardcoded in this component.
+ */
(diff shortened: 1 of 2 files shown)


danieldides (approved):
LGTM. A quick [search](https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/sourcegraph/handbook%24+file:%5C/cloud&patternType=standard&sm=1&groupBy=repo) seems to suggest this will only hit pages we want. 👍

michaellzc (approved):
lgtm

unknwon (commented):
Nice and cool!

PR #5797, opened 2022-12-13
add plg sequence diagram for cloud

michaellzc:


bobheadxi (approved):
Wish we had mermaid in handbook 🥲

PR #5850, opened 2022-12-20
update entitle access workflow

michaellzc:
workaround because requesting specific permission doesn't work

also, this way requires less click


unknwon (commented):
LGTM 👍

PR #5910, opened 2023-01-03
managed smtp: update instance handoff, note that managed SMTP is enabled by default

bobheadxi:
1. Document how password reset links are sent
2. Update managed SMTP page to note that it is enabled on all instances by default


danieldides (approved):

daxmc99 (approved):

filiphaftek (commented):
Thank you!

PR #5957, opened 2023-01-06
Add resources file to cse-k8s

gabtorre:


DaedalusG (approved):
Nice to document this for newer admins -- and also just to keep track of which overlay we're using 😆

PR #6010, opened 2023-01-11
Updates to scale testing deploy instructions

chwarwick:
Adds a reminder to connect to the correct cluster as well as adding the namespace when getting the history.


kalanchan (approved):
wooh, thanks for updating this!

PR #6034, opened 2023-01-16
Remove the reset-password instructions

rafax:
We no longer need to do it manually.


rafax:
@bobheadxi is there a way to check if the reset-password email was sent (in UI? in DB? in our email provider UI)?

filiphaftek (approved):
thx

rafax:
Added a way to check if email was sent.

PR #6038, opened 2023-01-16
Update POC licensing guidelines

justdueck:
Updates licensing guidelines to recommend that a short-term license key be provided to new trial admins during the deployment and configuration stage prior to a full POC license being issued. 

This will ensure that admins for enterprise POCs are able to fully configure and test their instance prior to POC rollout without encountering any gating issues.


ryphil (approved):

amenne (commented):
I'm curious why this is the case. Seems like added friction to ask them to apply two licenses, is it? I also don't see that we clarify free vs ent licenses, should we?

justdueck:
@amenne I don't think this adds much, if any, friction to an admin during the deployment process. Our license keys are very easy to apply and manage from an admin perspective. I think the extra step(s) that it takes to do an "admin" key for initial setup and then the full POC key, later on, is worth it to mitigate any license key-related risk that may come up during the deployment process for a prospect. 

Agreed that we need more clarity on Free vs Ent licenses. Merging this PR so those changes can be updated.

PR #6055, opened 2023-01-19
Updated some of the wording and reasoning

nickyvm:
I wanted to make it clear that we will not do email changes unless there is a very good reason for doing so however requesting an alias is far simpler


kemperhamilton (commented):
Looks good, I'm aligned and will merge this!

PR #6070, opened 2023-01-22
Add ICP to strat

malomarrec:
This is not meant to be comprehensive and is just a quick fix because our ICP was not in the strategy and hard to reference. I added a one-liner about adoption strategy vs ICP as it was needed for clarity, but I expect this whole page to be revamped as we enter FY24 so this is not at all meant to be the end state.


sqs (approved):

sqs:
I'm deleting this section in https://github.com/sourcegraph/handbook/pull/6133 because the linked MSEM doc's "Win in big companies with Big Code" section covers this. If you think not, let me know.

malomarrec:
👍

PR #6155, opened 2023-02-03
Update licensing page to clarify true-up on MAU plans

ryphil:


kopancek (approved):
Looks correct to me

pjlast (approved):
Thanks!

PR #4884, opened 2022-09-21
Propose Integrations team support rotation

philipp-spiess:
We've talked about adding a support rotation during our Barcelona offsite ([doc here](https://docs.google.com/document/d/1oHt7RvKnt9LqMCYikRif2H158fcxYPnaeGsoywfpcQQ/edit#)). This is my attempt to put this into writing based on the excellent Security support rotation document that was initially authored by @andreeleuterio (🙇).

Let's discuss any specifics inline. For now the scope is very small (since we only expect the member to respond to and triage support requests) but we can easily expand the responsibility to other functions (IIRC @ryankscott had some ideas about that).

I also suggest that we are re-evaluating the effectivity of this in about 2 months from now (I've only outlined the schedule until Nov 21 for now).

Defining a schedule as a markdown table is also not ideal lol but we can change the implementation of this later on.

PR #6306, opened 2023-02-28
Adding support rotation

scalabilitysolved:
Initial draft at a lightweight process for handling incoming support and requests for help for Code Exploration.

### DIFF
--- a/content/departments/engineering/teams/code-exploration/index.md
+++ b/content/departments/engineering/teams/code-exploration/index.md
@@ -67,7 +67,7 @@ As the largest frontend team, we also recognize our responsibility to be a stewa
 
 ### Support
 
-We run a weekly support rotation, help can be reached by tagging `@code-exploration-support` within Slack.
+We run a weekly support rotation, help can be reached by tagging `@code-exploration-support` within Slack. The engineer on rotation is expected to comply with these responsibilities only **_within their usual working hours_**. There is no on-call, 24x7-watch expectation on these tasks. The async nature of Sourcegraph allows us to properly support the organization without having to be in front of the computer all day.
 
 #### Responsibilities of the support rotation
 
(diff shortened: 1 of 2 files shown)


philipp-spiess (approved):
This is awesome stuff! Should we add some clarification around the time commitment of this support rotation? 

I wrote a similar proposal for the integration team before it fell apart (😢) and one thing I wanted to make clear is that the expectation is strictly to be doing the triaging during the usual work hours. 

Here's the page for integration for reference: https://github.com/sourcegraph/handbook/pull/4884/files#diff-7efe0e3a4e015b7d3b0c5913e62c457ef48bc4ce86bca1e04ce05a9d3de17551R11

scalabilitysolved:
> This is awesome stuff! Should we add some clarification around the time commitment of this support rotation?
> 
> I wrote a similar proposal for the integration team before it fell apart (😢) and 
> 
> Here's the page for integration for reference: https://github.com/sourcegraph/handbook/pull/4884/files#diff-7efe0e3a4e015b7d3b0c5913e62c457ef48bc4ce86bca1e04ce05a9d3de17551R11

By time commitment you mean the expected amount of time you'd see devoted to support during a week?  If it's that, I think it could be hard to quantify, I assume some weeks it'll be almost nothing and others quite busy.  What would you suggest?

philipp-spiess:
> By time commitment you mean the expected amount of time you'd see devoted to support during a week? If it's that, I think it could be hard to quantify, I assume some weeks it'll be almost nothing and others quite busy. What would you suggest?

I was thinking of two things:

- Limiting the support to working hours only. I saw that you've added that already.
- Mentioning something about the actual time this will take during the working hours. I agree that we don't know that yet but maybe something like "try to set aside 30% of your time this week for eventual support tasks" could still be helpful. We could use this as a guideline when planing things so that it's clear that we won't have 100% availability during our support week.

taras-yemets (approved):
Great! Thank you, @scalabilitysolved!

valerybugakov (approved):

PR #6377, opened 2023-03-07
Replace TripActions with Navan

BolajiOlajide:
[Context](https://sourcegraph.slack.com/archives/CQ44Y7F4G/p1678224435738109)

Since the change of name from TripActions to Navan, we ought to update the handbook to reflect this change.

[_Created by Sourcegraph batch change `BolajiOlajide/replace-navan`._](https://sourcegraph.sourcegraph.com/users/BolajiOlajide/batch-changes/replace-navan)


BolajiOlajide:
It works!

PR #6383, opened 2023-03-08
Update category usage to be consistent

ErikaRS:


ErikaRS:
Carly, I'm asking for a review on this one because I want to make sure that I'm not misreading what the intent is on wording here.

PR #6442, opened 2023-03-15
Make IC levels proper text vs image

eseliger:
Screenshot 2023-03-15 at 03 33 25@2x

This makes it searchable and more accessible to screen readers.


courier-new (approved):
I was just going to do this myself, thought to check open PRs first just in case and here you are. ❤️ So much better!

mrnugget (approved):
Impressive! Thank you!

PR #6468, opened 2023-03-17
Set width variable based on character size

courier-new:
Sets a `--container-width` custom property based on character width to apply to the parent container of all the career level framework pages with tables. This width makes it such that 4 full columns can fit on the page at any given time. If a table contains more than 4 columns, the container is wide enough for to peak the 5th column, revealing there's horizontal overflow. For example:

4 Columns | 5+ Columns
-----------|-------------
image | image

### DIFF
--- a/content/departments/engineering/design/career-development.md
+++ b/content/departments/engineering/design/career-development.md
@@ -17,7 +17,7 @@ It’s important to understand that what is listed in the level descriptions are
 
 <style>
   .container {
-    --width: 1300px;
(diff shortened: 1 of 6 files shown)


kopancek (approved):
Awesome, thanks! Not sure if you want to straight merge this to main, as your change is more comprehensive than mine.

courier-new:
Oops just read your edited comment 😂 It's fine, doesn't matter too much 🙂

PR #6472, opened 2023-03-17
Update use cases with direct links

ErikaRS:
Link to the use cases directly since the strategy page currently doesn't


ErikaRS:
Links aren't resolving correctly (yaml magic?) plus looking at the edit history of the strategy page, I realized I don't know what's going on with use cases.

PR #6563, opened 2023-03-31
m1-m2.md

milespatton:
Updated MacBook Pro models from M1 to M2


nickyvm (approved):
Thank you for taking the initiative on this!

PR #6564, opened 2023-03-31
cloud: add instructions for restarting v2 frontends

danieldides:


michaellzc (commented):
let's just move this section to https://handbook.sourcegraph.com/departments/cloud/#faq-how-do-i-restart-the-frontend-after-changing-the-site-config

PR #6610, opened 2023-04-04
cloud: update FAQ

unknwon:
I think this paragraph is no longer relevant in the SOAP world?


bobheadxi (commented):
~I thought the removed line pertains to SOAP?~ Ah I see - we should replace it with a mention of actual SOAP in that case

michaellzc (approved):

unknwon:
@bobheadxi - pushed another commit to add a mention of SOAP, PTAL!

PR #6632, opened 2023-04-06
update PM role pages

sqs:


sqs:
@malomarrec I updated the PM role description to match my intent. WDYT? Also, regarding the [`product_manager_engineering_manager_responsibilities.md`](content/departments/engineering/product/roles/product_manager_engineering_manager_responsibilities.md) page, is it actually helpful? Or is the short list in the PM roles page `You **own an area of the product**, which means you: ...` sufficient?

cc @ryphil @taylorsperry

PR #6673, opened 2023-04-10
modify-laptop-info.md

milespatton:
Removed computer sizes and updated links from CDW to Apple website


nickyvm (approved):
Looks good. Thank you

PR #6719, opened 2023-04-13
Finance Collections Policy

SamU713:
This is the official sourcegraph policy on collections for out accounts receivable balances


SamU713 (commented):
Corrected Link error

SamU713 (commented):
Removed link

PR #6753, opened 2023-04-17
Updated escalation engineer rotation to new format plus small edits to the content

johnwesonga:
Added information about the updated rotation format plus made some small edits.

### DIFF
--- a/content/departments/engineering/dev/process/escalation-engineer-rotation.md
+++ b/content/departments/engineering/dev/process/escalation-engineer-rotation.md
@@ -17,10 +17,11 @@ Discuss escalation engineering rotation in #escalation-engineering.
 | 2022-10-10 to 2022-11-07 | [@tbliu98](https://github.com/tbliu98), [Thorsten Ball](https://github.com/mrnugget)                                                          |
 | 2022-11-08 to 2022-12-05 | [@jhchabran](https://github.com/jhchabran)                                                                                                    |
 | 2022-12-06 to 2023-01-03 | [William Bezuidenhout](https://github.com/burmudar)                                                                                           |
-| 2022-01-03 to 2023-02-03 | [Camden Cheek](https://github.com/camdencheek), [Julie Tibshirani](https://github.com/jtibshirani) ,[Manuel Ucles](https://github.com/mucles) |
-| 2022-02-03 to 2023-03-03 | [Beatrix Woo](https://github.com/abeatrix), [Jacob Pleiness](https://github.com/jdpleiness) ,[Manuel Ucles](https://github.com/mucles)        |
-| 2022-03-03 to 2023-04-05 | ?                                                                                                                                             |
-| 2022-04-05 to 2023-05-05 | [David Veszelovszki](https://github.com/vdavid), [Milan Freml](https://github.com/kopancek), [Idan Varsano](https://github.com/varsanojidan)  |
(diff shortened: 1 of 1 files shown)


content/departments/engineering/dev/process/escalation-engineer-rotation.md
```diff
@@ -69,12 +70,15 @@ As an escalation engineer, you're effectively becoming the DRI (and tech lead) o
   - If you find yourself to not be able to lead, you are responsible for finding a person to replace you.
   - You keep the support request alive, you raise the alarm if it's stagnating or ask to close the support request if the context changes.
 
+Escalation Engineers will also **work on low effort "paper cuts" projects, typically scoped to 2-3 weeks maximum**. These projects will primarily consist of product gaps identified and prioritized by GTM with input from Engingeering & PM. Customer ad hoc issues still remain the P0 for the EE, with the projects as a P1.
```
vdavid, line 73: Until now, escalation engineering was something that people did on the side.
Does this addition mean that it will become a full-time thing? If there is a backlog of papercuts to work on, then that allows me to fill my workday with working on them.

I think with this change, clear guidance is needed on whether this should be considered a full-time role. Or if we _don't_ want to give clear guidance on this because it might vary case by case, we should be explicit about that.

Also, what's the source of these papercuts? If there's an \#escalation-engineering label or GitHub board, it'd be nice to link it here; if there isn't, we should point at \#feedback-dogfood and ask for proactivity in this area.
My suggestion is to ensure there is a **label**, mention it here, plus figure out a process to get the label added to papercuts.

johnwesonga: - **escalation engineering was something that people did on the side**. - my understanding is that during rotation the engineers are solely focused on addressing rotation related issues and are not expected to undertake other projects, so no split focus.

- **Does this addition mean that it will become a full-time thing?** - it is a 4 week commitment with the expectation that engineers will not be undertaking other job fair related projects, related to my response above.

- **what's the source of these papercuts?** GTM with input from Eng/PM has compiled this [list](https://docs.google.com/spreadsheets/d/152N7rUj24acdrPeg4HrLEpA4VERG_RJCQsvatEGPhWs/edit#gid=1009484269) to start with, I'll create a GH project board so that EEs can see what's on the board early and pick up to do when the rotation is light on issues. The paper cuts are P1 in terms of focus for the EEs.

vdavid: Thanks for the clarification, @johnwesonga!

Comments to each point:

1. I have a different understanding of how things worked _before_ this change. Without the papercuts, the only responsibility of EEs is to drive the open issues. In my short experience, it happens that _all_ open issues are waiting for customers, so there's not much else to do. And it often only takes about 1h/day to push everything forward as much as reasonable/possible, with 5 issues open. So dedicated EE doesn't feel reasonable—but of course, adding the papercuts to the mix changes this!

2. Thanks! My interpretation is a "yes".

3. This list is awesome; thank you! The only thing is that engineers don't usually have Salesforce access (I don't), so we can't open any of the links. I think the reason is that Salesforce seats are fairly expensive. I don't know how Salesforce pricing works, but if there's an option for a temporary monthly plan, then the two EEs can get it for their time. But then we'll need to add this to the process to give and revoke Salesforce access to EEs.

johnwesonga: 3) GH Issues Board with paper cuts https://github.com/orgs/sourcegraph/projects/336/views/1

johnwesonga:
Updates to escalation eng content

PR #6795, opened 2023-04-20
Add the design board and how to create an issue to how we work

rrhyne:


toolmantim (approved):
Great! I was making sure the word “design board” is in there for search, and it is! Thanks Rob.

PR #6807, opened 2023-04-21
Add link to CE message and re-genarate dashboard

filiphaftek:
from [slack discussion](https://sourcegraph.slack.com/archives/C03JR7S7KRP/p1682026360679719).

Should help CE to find proper instance and actions easier.

### DIFF
--- a/content/departments/cloud/technical-docs/v2.0/creation_process.md
+++ b/content/departments/cloud/technical-docs/v2.0/creation_process.md
@@ -67,7 +67,7 @@ In the GitHub issue, tag the assigned CE/AE the instance is ready with the follo
 Hi,
 
 The instance has been provisioned and set password email has been sent to the mentioned customer admin.
-For instance operations, please go to [production instances](http://go/cloud-ops).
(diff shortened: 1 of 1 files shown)


content/departments/cloud/technical-docs/v2.0/creation_process.md
```diff
@@ -276,6 +277,13 @@ cd terraform/stacks/tfc
 terraform init && terraform apply -auto-approve
 ```
 
+Add new instance to dashboard
```
filiphaftek, line 280: @michaellzc as this step is not automated afaik, I propose we do it always as part of create - dunno if we should move it to wrap up? b/c GithubAction is not doing this.

michaellzc (approved):

content/departments/cloud/technical-docs/v2.0/creation_process.md
michaellzc: manual is fine

or we can use gitub actions to regneerate this one on a schedule

PR #6980, opened 2023-05-24
cloud/v2: best-effort on-prem-to-cloud data migration refresh

bobheadxi:
A simple best-effort refresh to the old on-prem data migrations guide for Cloud v2 that I've been promising for a while. Where an equivalent in Cloud v2 exists, I've updated the guide for the equivalent - where one doesn't, I've left a `TODO` indicating as such. Pending further customer interest/requirement for a data migration there are no plans at the moment to fill in the gaps.

This process will be owned by @sourcegraph/implementation-eng but Cloud is still committed to supporting the first few migrations if they ever come up, and since the process pertains considerably to Cloud tooling and infrastructure I think the guide should live in the `/cloud` handbook pages.

Part of https://github.com/sourcegraph/customer/issues/1525


unknwon (approved):

content/departments/cloud/technical-docs/v2.0/onprem_data_migration.md
```diff
@@ -0,0 +1,249 @@
+# On-prem data migration to Cloud v2
+
+This process describes the current state of how to do a full data migration of an on-prem instance to a [Cloud v2](./index.md) instance.
+On-prem-to-Cloud data migrations are currently owned by [Implementation Engineering](../../../technical-success/ie/index.md), but the process is documented in Cloud as it pertains to Cloud infrastructure.
+
+> WARNING: This process is still a work-in-progress. For more details, see [RFC 760](https://docs.google.com/document/d/1IAgXmv2TbtU_rWXtph-KFc3qbqswREIAxO1rGALuoMQ/edit#) and the [tracking issue](https://github.com/sourcegraph/customer/issues/1525), or cc @bobheadxi (Robert Lin).
```
rafax, line 3: Maybe move this to the top of the page?

content/departments/cloud/technical-docs/v2.0/onprem_data_migration.md
```diff
@@ -0,0 +1,249 @@
...
+> This page was adapted only on a best-effort basis for Cloud v2 from [the legacy MI v1.1 page](../v1.1/index.md) - it has currently not yet been thoroughly tested for Cloud v2.
+
+## Requirements
+
+To qualify for a data migration, the customer must:
+
+- have a Sourcegraph instance on v3.20.0 or later
```
rafax: I'd mention that this is a MVU limitation

content/departments/cloud/technical-docs/v2.0/onprem_data_migration.md
```diff
@@ -0,0 +1,249 @@
...
+  - note: where possible, strongly encourage the customer to upgrade to their on-prem instance to the latest version of Sourcegraph first.
+- use databases on Postgres 11 or later
```
rafax: We don't support 11 https://docs.sourcegraph.com/admin/postgres#version-requirements

rafax (approved):
Thanks!

content/departments/cloud/technical-docs/v2.0/onprem_data_migration.md
```diff
@@ -0,0 +1,249 @@
...
+    }
+  ]
+}
+```
+
+### Create snapshot contents
+
+#### Databases
+
+The customer should first be asked to create `pg_dump` exports of their Sourcegraph databases. Template commands can be generated with `src snapshot databases` for various configurations:
```
loujar: Shouldn't we quiesce the Sourcegraph services prior to taking the pg_dump snapshots? At least in an ideal scenario where we have some flexibility in taking the instance down temporarily.

content/departments/cloud/technical-docs/v2.0/onprem_data_migration.md
bobheadxi: At some point we did support 11, so it's likely there are 3.20-ish instances that are on it. IIRC it was the upgrade _to_ 11 that caused issues, hence the requirement here - happy to bump that though

content/departments/cloud/technical-docs/v2.0/onprem_data_migration.md
bobheadxi: IMO no - `pg_dump` is designed to be usable while the database is in use: https://www.postgresql.org/docs/current/app-pgdump.html

> It makes consistent backups even if the database is being used concurrently. pg_dump does not block other users accessing the database (readers or writers).

Instead, as part of this process we ask the customer to configure a notice to let their users know that any actions taken after the point of the dump will not remain consistent on their new Cloud instance

content/departments/cloud/technical-docs/v2.0/onprem_data_migration.md
bobheadxi: Bumped the requirement regardless 👍

content/departments/cloud/technical-docs/v2.0/onprem_data_migration.md
bobheadxi: Added a section to the doc here that mentions this!

PR #7018, opened 2023-05-30
Secure development policy review

andreeleuterio:
@doragrgic the only bit I see as possibly problematic is [this](https://handbook.sourcegraph.com/company-info-and-process/policies/secure-development-policy/#system-change-control-procedures):

> Changes to systems within the development lifecycle shall be controlled by the use of formal change control procedures. Change control procedures and requirements are described in the Sourcegraph Operations Security Policy.

I could not find a Sourcegraph Operations Security Policy to confirm those controls. I updated the policy to remove this.


doragrgic (commented):
Looks good with the changes. Yes, we do not have such a policy. Thanks for rewording it.

dcomas (approved):

PR #7057, opened 2023-06-01
Rename channel name

nelsonjr:
* Renamed #announce-people-team to #announce-people-talent-team to match channel name.


madisongclark (approved):
Thanks for catching this, @nelsonjr

PR #7064, opened 2023-06-05
Update with rotation starting June 5

gl-srgr:
I'm not sure if we need to include regional coverage or if we have distinct regional coverage this rotation. Let me know in PR if you want to add anything related to regional coverage / time zone coverage.


marekweb:
Is the rotation intended to be 4 weeks or 1 calendar month? The handbook page defines it as 4 weeks which would put the end date on June 30. But I see that we've actually done calendar months in past rotations.

If we're doing calendar months then let's update the handbook page to reflect that.  Although I'd say that defining it as a number of full weeks would be a clearer and more predictable duration. Can you weigh in @johnwesonga?

johnwesonga:
@marekweb the original EE rotation had 8 calendar weeks, typically started on the 5th of one month and ended on the 5th 3rd month e.g. Jan/05 - Mar/05, we shortened this to 4 calendar weeks which would be Jan/05 - Feb/05, I think the rationale behind it was everyone does the same duration = 4 weeks, versus using calendar months where some months might be shorter e.g. February. I hope this makes sense

marekweb (approved):

PR #7083, opened 2023-06-07
cloud: update instructions for determining Project ID

danieldides:
These instructions were out of date and referred to MI 1.1. The process is different (and easier!) now in v2.


bobheadxi (approved):

unknwon (commented):
Thank you!

PR #7264, opened 2023-06-29
init grpc doc

mucles:
started a doc page for grpc related things

### DIFF
--- a/content/departments/engineering/dev/tools/grpc.md
+++ b/content/departments/engineering/dev/tools/grpc.md
@@ -1,27 +1,8 @@
 # gRPC
 
-## Introduction
-
 This document provides an overview and documentation for Sourcegraph-Specific gRPC usage, including information about gRPC services, monitoring, common issues, and additional resources. The purpose of this document is to assist developers in understanding and working with gRPC.
 
-## Table of Contents
-
-1. [gRPC Services](#grpc-services)
-2. [Monitoring](#monitoring)
-3. [Known Issues](#known-Issues)
-4. [Additional Resources](#additional-resources)
-
-## gRPC Services
-
-### Gitserver
-
-[gitserver](https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/tree/cmd/gitserver/README.md) has been migrated to use gRPC. The see the following for more information: [gRPC: migrate gitserver](https://github.com/sourcegraph/sourcegraph/issues/46522)
-
-### Frontend
-
-[frontend](https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/tree/cmd/frontend/README.md) is still in progress of being migrated to use gRPC yet. The see the following for more information: [gRPC: migrate internal frontend APIs](https://github.com/sourcegraph/sourcegraph/issues/46525)
-
-## Monitoring
+## GRPC Monitoring Guide
 
 to view the gRPC metrics, go to the built-in Grafana dashboard. The goal of these metrics is to identify specific types of errors that we are aware of, which originate from the grpc/protobuf libraries we employ. Such errors often hint at potential configuration issues or bugs related to our grpc-specific application logic. By implementing these additions, we aim to gain better insights into these errors and investigate any necessary adjustments or fixes.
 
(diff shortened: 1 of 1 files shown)


content/departments/engineering/dev/tools/grpc.md
```diff
@@ -0,0 +1,76 @@
...
+[gitserver](https://sourcegraph.com/-/debug/grafana/d/gitserver/git-server?orgId=1) for example. Within that dashboard, there are two sections with panels for gRPC metrics called `GRPC server metrics` and `GRPC "interanl errors" metrics`.
+
+There is also a filter to only view metrics for a specific rpc method at the top of the datchboard. e.g. https://sourcegraph.com/-/debug/grafana/d/gitserver/git-server?orgId=1&var-alert_level=All&var-shard=All&var-method=Archive
+
+### Metrics
+
+#### GRPC server metrics
+
+- `Request rate per-method over 2m` - This is a good indicator of wether or not the rpc method is being used if at all.
+- `Error percentage per-method over 2m` - Captures server side errors that are not related to grpc itself. For example, if the rpc method is called and later given a cancellation signal, then it is expected that the rpc method will fail. In this case, the error percentage will be high, but it is not a cause for concern because the failures are unrelated to grpc itself.
```
ggilmore: > ... Captures server side errors that are not related to grpc itself.

This is not quite right. The nuance here is that every gRPC operation returns a [status code](https://grpc.github.io/grpc/core/md_doc_statuscodes.html) (which is the same core idea as the [HTTP status codes](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status), just different). 

These charts plot responses with only "status.OK" against the total set of responses. Some core things to think about here are:

- **Both normal "application logic" errors and errors that _might_ be related to the specifics of the gRPC implementation are captured in this chart**. 
- There are many reasons for a method to fail that's unrelated to the exact transport mechanism used (this repository doesn't exist., etc.)
- Some failure modes are expected. For instance, [status.DeadlineExceeded / or status.Cancelled](https://grpc.github.io/grpc/core/md_doc_statuscodes.html) is frequently returned when the caller intentionally cancels the operation. However, this could be an entirely intentional part of our application logic (example: there is no need to continue streaming the output of a git archive call if the user cancelled the search or navigated away from the blob page). 
   	- See https://sourcegraph.com/-/debug/grafana/d/gitserver/git-server?viewPanel=100711&orgId=1&var-alert_level=All&var-shard=All&var-method=Archive and https://sourcegraph.com/-/debug/grafana/d/gitserver/git-server?viewPanel=100740&orgId=1&var-alert_level=All&var-shard=All&var-method=Archive
-

content/departments/engineering/dev/tools/grpc.md
```diff
@@ -0,0 +1,76 @@
...
+
+[gitserver](https://sourcegraph.com/-/debug/grafana/d/gitserver/git-server?orgId=1) for example. Within that dashboard, there are two sections with panels for gRPC metrics called `GRPC server metrics` and `GRPC "interanl errors" metrics`.
+
+There is also a filter to only view metrics for a specific rpc method at the top of the datchboard. e.g. https://sourcegraph.com/-/debug/grafana/d/gitserver/git-server?orgId=1&var-alert_level=All&var-shard=All&var-method=Archive
+
+### Metrics
+
+#### GRPC server metrics
+
+- `Request rate per-method over 2m` - This is a good indicator of wether or not the rpc method is being used if at all.
```
ggilmore: It would be useful to link to the sourcegraph.com dashboards as an example here: https://sourcegraph.com/-/debug/grafana/d/gitserver/git-server?viewPanel=100701&orgId=1&var-alert_level=All&var-shard=All

content/departments/engineering/dev/tools/grpc.md
```diff
@@ -0,0 +1,76 @@
...
+- `Error percentage per-method over 2m` - Captures server side errors that are not related to grpc itself. For example, if the rpc method is called and later given a cancellation signal, then it is expected that the rpc method will fail. In this case, the error percentage will be high, but it is not a cause for concern because the failures are unrelated to grpc itself.
+
+the following helps you track the 99th, 90th, 75th percentile response time for each rpc method over a 2-minute interval. By monitoring this metric, you can identify any performance issues or outliers that may affect user experience or indicate areas for optimization.
+
+- `99th percentile response time per method over 2m`
```
ggilmore: Good, but please link Grafana dashboards here

content/departments/engineering/dev/tools/grpc.md
```diff
@@ -0,0 +1,76 @@
...
+- `90th percentile response time per method over 2m`
+- `75th percentile response time per method over 2m`
+
+#### GRPC "internal errors" metrics
+
+- `Client baseline error percentage per-method over 2m` - captures the percentage of errors that are related to grpc itself. For example, if the rpc method is called and later given a cancellation signal, then it is expected that the rpc method will fail. In this case, the error percentage will be high, but it is not a cause for concern because the failures are unrelated to grpc itself.
```
ggilmore: This is incorrect, this it's not only "internal errors" - it's all errrors (whether or not we know that they originate from the gRPC library)

content/departments/engineering/dev/tools/grpc.md
```diff
@@ -0,0 +1,76 @@
...
+- `Client-observed gRPC internal error percentage per-method over 2m` - This capture the percentage of errors that we know originate from the grpc / protobuf libraries we use
+- `Client-observed gRPC internal error response code rate per-method over 2m` - This gives you a generally idea of how often a specific grpc status code is returned.
+
+Here are the posssible [grpc status codes](https://grpc.github.io/grpc/core/md_doc_statuscodes.html) that can be returned by a rpc method.
+
+#### More on internal errors
+
+The GRPC "internal errors" metrics flags specific types of errors that we can attribute to the grpc/protobuf libraries we use. Usually, these errors might hint at some configuration or application logic issues specific to grpc that may need further investigation.
+
+This errors are currently being captured through the use of a grpc feature called [Interceptors](https://github.com/grpc/grpc-go/blob/master/examples/features/interceptor/README.md).
```
ggilmore: Don't think that the use of interecptors specifically is releavant here. I think it's sufficent to say that we check to see if an "internal" error is returned after every message sent or received from the client or server.

content/departments/engineering/dev/tools/grpc.md
```diff
@@ -0,0 +1,76 @@
...
+The go-grpc library has a tendency to prefix most gRPC client/server errors with "grpc: ...". Our new interceptors are designed to detect this string pattern.
+
+Some of the limitations of this approach are:
+
+- The error strings used by https://github.com/grpc/grpc-go could be modified in the future.
+- Not all errors from https://github.com/grpc/grpc-go are necessarily transmitted over the network.
+- It's possible for someone to inadvertently create an application-level error that starts with "grpc: ...".
+- Errors originating from the library may not always be the library's fault, such as when encountering an application bug like sending a 5GB symbols response (see [gRPC 4mb message size limit](#grpc-4mb-message-size-limit)).
+
+See the this [sourcegraph search](https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/internal/grpc/internalerrs/common.go?L160-166) to see the current list of errors that are being captured as "internal errors"
```
ggilmore: This should be a permalink, as the line numbers can change as when the file is modified.

ggilmore (commented):
first pass, looks okay so far

content/departments/engineering/dev/tools/grpc.md
```diff
@@ -0,0 +1,76 @@
...
+
+Despite these limitations, implementing such checks (which may need to be modified or removed later) enhance our monitoring capabilities as we further develop the gRPC implementation.
```
ggilmore, line 99: I would also describe the fact that all of these internal errors are logged with the special log scope that contains `gRPC.internal.error.reporter` - this allows to filter the logs in GCP for just these errors. 

See [go/prod-logs-grpc-errors](https://cloudlogging.app.goo.gl/jwMrrNd777ZzTErt7)

ggilmore (approved):

ggilmore:
Nice! We'll probably need to iterate on it over time given new developments and feedback, but this is good enough to merge now

PR #7289, opened 2023-07-06
update post creation msg

michaellzc:


content/departments/cloud/technical-docs/v2.0/creation_process.md
```diff
@@ -68,6 +68,8 @@ Hi,
 
 The instance has been provisioned and set password email has been sent to the mentioned customer admin.
 For instance operations, please go to http://go/cloud-ops.
+
+Cody has been setup by default, please instruct customer admin to configure `code.enabled=true` in site config.
```
unknwon, line 72: This is unfortunate due to the nature of our site config, but I think using the correct format `"cody.enabled": true` is better to avoid any confusion. WDYT?

PR #7335, opened 2023-07-11
cloud/v2: updated onprem to cloud data migration process

bobheadxi:
Requires tooling addition https://github.com/sourcegraph/controller/pull/853


bobheadxi:
This is untested, but marking it as ready for review + merge as I will be OOO next week

PR #7433, opened 2023-07-21
clarifying expectations for performing Cloud v2 migrations

loujar:
Adding a couple of edits to ensure TAs request cloud migrations at the correct time in the cloud provisioning process.


anorrish (approved):
LGTM!

PR #7503, opened 2023-07-30
Weekly reports for ship

nelsonjr:


content/departments/engineering/ship/ship_okr_square.png
unknwon, line 1: Looks like this image is not used in any part of this PR?

PR #7566, opened 2023-08-10
Add capitalization rules for product integrations

toolmantim:
We want to feel like like first-class integrations when we’re adding UI to other people’s products. This adds a note to our capitalization rules to allow and encourage this.


sqs (approved):

rrhyne (approved):
Good add.

PR #7662, opened 2023-08-18
Update team.yml

Shannonking456:
Updating bio in handbook


Shannonking456:
Updating my bio in the handbook.

PR #7690, opened 2023-08-24
Fix typo in Kevin Chen's bio

vdavid:
Just stumbled upon it :)


chenkc805 (approved):
updated the "with" that was changed :)

PR #7710, opened 2023-08-30
Update license_keys.md

ghost:
Added steps to perform best practices when creating a user for a license


simonintech:
Hi @n8maynard but is there a requirement to add an email to a user account which will be associated with a subscription? So my push back would be that this change suggests that we should be adding a verified link to a user account.

PR #7731, opened 2023-09-01
cody-gateway: update redash docs

bobheadxi:


camdencheek (approved):
Ah, I see now, thanks!

PR #7738, opened 2023-09-04
chore: add important special block style

unknwon:
###### Important

Before:
CleanShot 2023-09-04 at 17 45 05@2x

After:

CleanShot 2023-09-04 at 19 17 10@2x

###### Warning

Before:
CleanShot 2023-09-04 at 19 17 49@2x

After:
CleanShot 2023-09-04 at 19 17 59@2x

###### Note

Before:
CleanShot 2023-09-04 at 19 18 37@2x

After:
CleanShot 2023-09-04 at 19 18 48@2x

### DIFF
--- a/src/lib/markdownToHtml.ts
+++ b/src/lib/markdownToHtml.ts
@@ -169,12 +169,12 @@ const rehypeExtractTitleFromH1: Plugin = () =>
 
 /**
  * Note blockquote syntax, originally from docsite. Any blockquote starting with
- * "> NOTE:" is converted to <aside class="note">...</aside>.
+ * "> [!NOTE]" is converted to <aside class="note">...</aside>.
  */
 const remarkSpecialNoteBlocks: Plugin<[], MdastRoot> = () =>
     function (tree) {
         for (const node of tree.children) {
-            if (isSpecialBlockquote(node, 'NOTE:')) {
+            if (isSpecialBlockquote(node, '[!NOTE]', 'Note:')) {
                 // TODO: This overwrites the `hProperties` to add the class
                 // name. Improve it by adding the class name while respecting
                 // existing `hProperties` or existing classes.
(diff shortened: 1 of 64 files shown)


michaellzc (commented):
can we reuse github flavored markdown? https://github.com/orgs/community/discussions/16925

unknwon:
> can we reuse github flavored markdown? https://github.com/orgs/community/discussions/16925

@michaellzc  You mean the markdown syntax part (`> [!IMPORTANT]`) and/or rendered style? (The former should be straightforward and only need global find-and-replace. Not sure how much work to make the latter work 😂

michaellzc:
> > can we reuse github flavored markdown? https://github.com/orgs/community/discussions/16925
> 
> @michaellzc You mean the markdown syntax part (`> [!IMPORTANT]`) and/or rendered style? (The former should be straightforward and only need global find-and-replace. Not sure how much work to make the latter work 😂

Ideally, we can support the `> [!IMPORTANT]` syntax.

🤔 it shouldn't be too hard, I think? When traversing the tree, whenever we see `[!IMPORTANT]` at the beginning of the codeblock, just remove the first node `> [!IMPORTANT]` and return the rest of stuff

https://unifiedjs.com/learn/recipe/remove-node/

styling doesn't really matter to me, as long as we can tell the difference between NOTE/IMPORTANT/WARNING

unknwon:
> Ideally, we can support the `> [!IMPORTANT]` syntax.

This part is actually easy IMO (supporting the syntax). Rendering exactly like the GitHub ones is unknown to me (CSS magic 😂 )

unknwon:
@michaellzc - alright, this PR has from small QoL improvement to a controversial one, see https://github.com/sourcegraph/handbook/pull/7738/commits/26417559f3fcebda872c4639615fb0b874e13a94 and updated screenshots.

src/lib/markdownToHtml.ts
unknwon, line 1: This is the file that really matters

michaellzc (approved):
nice!

unknwon:
OK, I hope people don't get mad at me 😛

michaellzc:
> OK, I hope people don't get mad at me 😛

lol, it's much easier to follow a more well-adopted standard instead of inventing our own. 

https://github.com/remarkjs/remark-gfm will catch up soon

filiphaftek (commented):
Awesome!

PR #7807, opened 2023-09-14
AWS accounts management under Ship

filiphaftek:
This PR describes how to:
- create AWS account
- login to Team AWS account via web console and cli/terraform
- delete AWS account

### DIFF
--- a/content/departments/engineering/ship/aws.md
+++ b/content/departments/engineering/ship/aws.md
@@ -1,0 +1,68 @@
+# AWS Cloud Management
+
+Sourcegraph AWS Organisation is managed by the Ship team. This includes:
+
+- creating new AWS accounts
+- OKTA group access assignments to AWS Accounts
+- billing alerting
+- deleting AWS accounts
+
+## Create AWS Account
+
+AWS Account are owned by team, which is responsible for requesting access and managing resources. To create a new AWS account:
+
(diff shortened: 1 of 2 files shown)


content/departments/engineering/ship/aws.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Create AWS Account
+
+AWS Account are owned by team, which is responsible for requesting access and managing resources. To create a new AWS account:
+
+1. Add new organisation unit and AWS account [here](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/infrastructure/-/blob/cloud/aws/variables.tf)
+
+2. Open Pull Request and ask for approval in #discuss-cloud-ops channel.
+
+> Note: default billing alert is set to 500$/m, if you require higher limit contact Finance Team and ask for approval in Pull Request.
```
filiphaftek, line 18: @TommyODonnell as discussed with @nelsonjr , we probably need to set minimal limit of spendings on account so it's easy to create new test account - could you pls decide on the amount? and how to request approval from Finance Team (if other than approving PR)?
We will setup billing alerting, but need to define limit when alert should be fired - in AWS this is even more important than in GCP, b/c some AWS services can be crazy expensive.

michaellzc (approved):

PR #7812, opened 2023-09-15
Design team updates

rrhyne:
This PR builds off of Tim's PR, but moves the entire structure of design over from /engineering to /product, alters the product page to display the three teams under product and link to the design section of the handbook, and makes additional cuts. 

Lastly, and notably, it revives issue templates which help standardize our process.


toolmantim (approved):
I originally removed the issues templates because I think they need to be revisited now we have PRDs and product. But we can do that in a follow up.

PR #8001, opened 2023-10-03
Update security-incident-response-process

mohammadualam:
This PR incorporates the security incident cheklist into our processes

### DIFF
--- a/content/departments/security/security-incident-response-process.md
+++ b/content/departments/security/security-incident-response-process.md
@@ -3,6 +3,7 @@
 The aim of this document is to define the key steps required as part of responding to a security incident, in line with Sourcegraph’s Security Incident Response Policy.
 
 ## Security Incident Lead
+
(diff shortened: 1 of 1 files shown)


content/departments/security/security-incident-response-process.md
```diff
@@ -2,6 +2,23 @@
...
+- Following the Security Incident Checklist.
+- Maintaining the list of action items and work streams, and assigning them to responders.
+- Validating action items have been completed.
+- Providing regular incident updates.
+- Ensuring the right teammates are working on the incident.
+- Handing communication within the team, with the company, and working with a messenger for customer-impacting incidents.
+
 ## Triaging a reported event and declaring an incident
 
 When triaging a security event, follow the documented [security event response process](https://github.com/sourcegraph/security-monitoring/blob/main/docs/security-event-handling.md).
```
willdollman, line 24: The following lines (they get collapsed in the diff) could use updating/removing - they refer to our old incident response template that we've replaced with incident.io.

mohammadualam: Got it, i will create a new PR with those changes just so that i can keep them separated for review, thanks!

PR #8087, opened 2023-10-18
Improve dotcom site admin docs

andreeleuterio:
Improvements suggested by @vdavid


vdavid (approved):
Thanks for all three changes! (the note, the `/access_request` mention and the entitle link)

PR #8133, opened 2023-10-26
Update guidelines for new HC approval process

dadlerj:


dadlerj:
There are more words here, but this will be a much simpler and quicker process than before :)

PR #8224, opened 2023-11-08
add AWS SSO Viewer for root account description

filiphaftek:
Description how to use root account in read-only mode.
Also changed the order as dev accounts should be mostly used.


michaellzc (approved):
can we add some pointer to the create account flow for users to check for potentially duplicated account?

PR #8231, opened 2023-11-09
Minor handbook fixes on team name and titles

airamare01:
This handbook update makes a few changes to the team name and Slack channel to represent what we have currently.


Rhia2 (approved):
Thank you boss

PR #8242, opened 2023-11-14
Fix JSON syntax in `locations.geojson`

StreakyCobra:
Remove the extra trailing comma that prevents the locations from showing on the map.

PS: Feel free to take this change in one of your own PR if that's easier from a legal/CLA perspective :-)


moedinger:
FYI, in addition to the extra trailing comma, there's another issue with the locations.geojson file: Three of the LatLong coordinates seem to be swapped.

In GeoJson format, longitude comes before latitude in a coordinate pair. In three cases, the latitude is outside the valid range of -90 to 90. For these cases, if you swap the lat and long, they are valid coordinates.

Here's the diff to show what I believe would be the correct version:

```text
@@ -126,8 +126,8 @@
             "geometry": {
                 "type": "Point",
                 "coordinates": [
-                    -122.9906622,
-                    49.2200776
+                    49.2200776,
+                    -122.9906622
                 ]
             }
         },
@@ -919,8 +919,8 @@
             "geometry": {
                 "type": "Point",
                 "coordinates": [
-                    -122.592757,
-                    45.512328
+                    45.512328,
+                    -122.592757
                 ]
             }
         },
@@ -932,8 +932,8 @@
             "geometry": {
                 "type": "Point",
                 "coordinates": [
-                    -104.99025,
-                    39.739235
+                    39.739235,
+                    -104.99025
                 ]
             }
         },
```

PR #8288, opened 2023-11-30
Improve steps to quickly provide a customer audit logs

daxmc99:
We may need to provide a customer audit logs quickly so we need docs that are easy and straightforward

Ctx: https://sourcegraph.slack.com/archives/C03JR7S7KRP/p1701301304366649?thread_ts=1701233124.526789&cid=C03JR7S7KRP


filiphaftek (commented):
nice!

PR #8336, opened 2023-12-11
Remove cse-aws instance mentions

gabtorre:
Remove mentions of the cse-aws instance after shutting down the instance.


morgangauth (approved):
LGTM thanks Gabe!

PR #8371, opened 2023-12-25
managed-services: incident response guide

bobheadxi:
This starts a page that collects useful infrastructure access links and guidance in the event of an MSP service outage caused by infrastructure issues.

It also outlines resource deletion and recovery in detail, with INC-263 as a specific example.

Later on we might want to aim to have parity with https://handbook.sourcegraph.com/departments/cloud/technical-docs/v2.0/disaster_recovery_process/ (looks like it focuses on zonal failures) - this isn't implemented in MSP yet

Closes https://github.com/sourcegraph/managed-services/issues/308


unknwon (approved):
Thank you!

PR #8452, opened 2024-01-17
Update slack channels to current for sales & TS.md

sldam:
Updating slack channels to match current channels for sales & TS teams


amenne (approved):
Made 2 small tweaks directly, lgtm!

PR #8509, opened 2024-01-31
Repo policies and controls

andreeleuterio:
As discussed in https://github.com/sourcegraph/security/issues/1025

### DIFF
--- a/content/departments/security/security-support-rotation.md
+++ b/content/departments/security/security-support-rotation.md
@@ -131,3 +131,8 @@ We are notified when a user submits request for vendor procurement.
 5. Review the attached documents with above guidelines in mind
 6. If a security portal link from the vendor is provided, request access and review information available there
 7. Click approve or deny based on your review
+
+
+### New repository alert
+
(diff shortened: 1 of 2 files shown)


content/departments/security/security-support-rotation.md
```diff
@@ -131,3 +131,7 @@ We are notified when a user submits request for vendor procurement.
 5. Review the attached documents with above guidelines in mind
 6. If a security portal link from the vendor is provided, request access and review information available there
 7. Click approve or deny based on your review
+
+### New repository alert
```
doragrgic, line 135: @andreeleuterio  Shall we make it more explicit in terms of what we expect users to do here when they create a new repo? Or did you mean we will have some sort of automation that will add the repos to the matrix?

doragrgic (approved):

content/departments/security/security-support-rotation.md
andreeleuterio: yup, the automation will add repos to the spreadsheet. I think it's best for us to add these checks on a few repos and see the pains before making it a support rota responsibility. So for now support rota would just add the repo to the list and categorize.

PR #8536, opened 2024-02-05
Update license management page to reflect the upcoming 5.3 changes

pjlast:
Updates to the license key management page to reflect changes after 5.3

Licensing plans are a lot more simplified after 5.3, and we'd like the docs to reflect this as well, so as part of this we're removing all the "Legacy" parts of the docs here, as they make things seem a lot more complicated than they need to be.

However, if we feel this information is absolutely required, I'd recommend adding a separate page with the legacy information to keep the complexity there.

### DIFF
--- a/content/departments/technical-success/ce/process/license_keys.md
+++ b/content/departments/technical-success/ce/process/license_keys.md
@@ -81,8 +81,6 @@ Below is an overview of the **current supported plans** and included features:
   </tr>
 </table>
 
-You can read more about legacy plans [below](#legacy-plans).
-
 #### Additional Options
 
 Apart from the plan, there are a few additional options that can be added to a license key.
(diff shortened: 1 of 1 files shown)


content/departments/technical-success/ce/process/license_keys.md
```diff
@@ -61,174 +61,57 @@ Below is an overview of the **current supported plans** and included features:
...
   </tr>
 </table>
 
-You can read more about legacy feature tags [below](#legacy-feature-tags).
-
-#### Legacy Plans
+#### Additional Options
 
-You may come across legacy licenses associated to legacy plans. Previously no-longer-used, legacy plans you may encounter are:
+Apart from the plan, there are a few additional options that can be added to a license key.
```
eseliger:
```suggestion
Apart from the plan, there are a few additional options that can be added to a license key, depending on the license type.
```

eseliger (approved):

amenne (approved):
Committed a couple changes to the PR directly, good to merge!

PR #8601, opened 2024-02-14
Delete old compensation & pay index page

jkoconne:
…ion-and-leveling/index.md


jkoconne:
Need to delete other links and re-delete

PR #8611, opened 2024-02-14
Update invoices.md to account for currency conversion fees

arafatkatze:
I faced a roughly 1.5% in conversion fee because I thought that I would be paid the salary in my currency which is Canadian Dollar. 

I got that impression from the existing description in this page. So now I am editing the page to help other people to avoid facing this.


desenesterling (changes requested):
- NOTE: If the the currency in your contract differs from your payment currency, Airbase convert your invoice amount into your desired payment currency. This may result in a favorable or unfavorable exchange rate. To mitigate this: I
  A. Retain a contract in the same currency as your payment currency. Updated employment contracts are handled by the People Team.
  B. Maintain an account in the same currency as your contract to avoid exchange rate instability.

arafatkatze:
image
@desenesterling I made the changes

amyjohnsonSG (approved):

desenesterling (approved):

PR #8649, opened 2024-02-27
Add guidance for using Linear during trial

varungandhi-src:
- [x] Importing issues from cody repo. Waiting on [Slack discussion](https://sourcegraph.slack.com/archives/C05AGQYD528/p1710984056320229)
- [x] Importing issues from the monorepo

### DIFF
--- a/content/departments/engineering/linear-trial.md
+++ b/content/departments/engineering/linear-trial.md
@@ -118,11 +118,13 @@ You can ask for help in `#wg-linear-trial` with these steps as well.
    is available for search.
 
    After this step, select 'Import users' and make sure that:
+
    - If a person has 'Map to Existing' but no associated account on
      the right side, then change it to 'Do not create'
    - Replace any 'Create with email' with 'Do not create'
 
    Once you click next, confirm on the next page that:
+
    - The number of user invites is zero
    - The number of completed issues matches up with the closed
      issues in the repo's GitHub issue tracker.
--- a/content/departments/engineering/linear-trial.md
+++ b/content/departments/engineering/linear-trial.md
@@ -4,10 +4,7 @@
 TODOs post adoption:
 - Remove trial from doc title
 - Combine with 'working with issues' doc
-- Collect feedback on labels vs label groups
-  - If we're using flat labels, then change the label
-    syntax to use `.` so that we can create new labels
-    inside the Linear UI.
+- Update guidance around label groups.
 - Update mentions of Eng Private team if we
   merge that with main Eng team. Maybe we still need
   a separate OSS team for OSS repos.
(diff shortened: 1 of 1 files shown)


content/departments/engineering/linear-trial.md
```diff
@@ -0,0 +1,234 @@
...
+      ([example link](https://linear.app/sourcegraph/settings/teams/ENG/labels)).
+
+5.  **Do explore the various Linear [features](https://linear.app/docs)
+    and [integrations](https://linear.app/sourcegraph/settings/integrations)**.
+
+## FAQ
+
+### Why are Eng PUBLIC and Eng Private separate?
+
+The Eng Private team is present because of the following
```
mmanela, line 181: Does this mean teams would need to have two boards? One for private and one for public work? Or is there a way to get a unified view?

content/departments/engineering/linear-trial.md
```diff
@@ -0,0 +1,234 @@
...
+This document describes recommended ways of using Linear
+during the trial period, without causing a disruption to
+team members who continue using GitHub
+and are not part of the trial.
+
+While the [Linear docs](https://linear.app/docs) are excellent,
+here is a very quick run down of the core mental model.
+
+- The workspace is divided into Teams.
+  - Settings related to workflow (issue statuses, estimates, cycles),
```
mmanela, line 22: Follow up on our conversation on cycles. Do we want to take a perspective for everyone here on a global cycle on or off? Given teams already use sprints today, might be worth putting a stake in the ground that we do 2 week cycles and see feedback?

content/departments/engineering/linear-trial.md
varungandhi-src: Views in Linear are not necessarily tied to Teams, so that shouldn't be a problem. I noticed that I mistakenly made the Graph team issues view to be scoped to the Eng PUBLIC team, but you can see the [new View](https://linear.app/sourcegraph/view/7b4400f7-6e98-4847-b628-6645871c0621) which contains issues from both Eng PUBLIC and Eng Private

content/departments/engineering/linear-trial.md
varungandhi-src: I don't know what the distribution for this is - can you share more details about how many teams are already using 2 week sprints?

mmanela: I know Source and Search Product (not platform) are doing 2 week sprints. Not sure about if anyone else

content/departments/engineering/linear-trial.md
mmanela: oh that is nice, views seem super flexible

content/departments/engineering/linear-trial.md
eseliger: We sometimes swift our cadence based on need and company events like merge to 1 week sprints where necessary, so we're generally using a flexible schedule, with a tendency to do 2 week sprints.

varungandhi-src: OK, given the direction we're going in with [RFC 916](https://docs.google.com/document/d/16mipKz0fc2Ki5NYXG8Rarb8b4KHh34t27P9Ao_m8e-o/edit#heading=h.9ebsu4sgujtv), I've just updated Eng PUBLIC and Eng Private teams to use 2 week cycles starting on Apr 1, will update this doc tomorrow.

PR #8650, opened 2024-02-27
Add a link to the new location of AP process

rafax:
Link to the page that contains the process.


varungandhi-src:
Thank you! I was like, I should do this, because of the search ranking, but I kept putting it off even though it is a tiny change

PR #8659, opened 2024-02-29
update patch upgrade process

michaellzc:
we now have the percentage based batch upgrade.

let's be more specific for the patch upgrade process


filiphaftek (approved):
Nice!

PR #8891, opened 2024-04-11
Replace references to release guild with releases team

craigfurman:
* Replace slack channel references to discuss-release-ship with discuss-releases
* Fix github team label for the releases team
* Replace references to release guild with the releases team

---

Please holler if any of this is wrong! Some more notes in the self-review 👇

### DIFF
--- a/data/guilds.yml
+++ b/data/guilds.yml
@@ -1,14 +1,1 @@
-release_guild:
-  title: Release Guild
-  leader: bolaji_olajide
-  leadership_sponsors: [quinn_slack]
-  members:
-    [
-      joe_chen,
-      camden_cheek,
-      keegan_carruthers-smith,
-      mohammad_umer_alam,
-      bolaji_olajide,
-      warren_gifford,
-      jacob_pleiness,
-    ]
+{}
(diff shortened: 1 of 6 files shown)


data/guilds.yml
```diff
@@ -1,14 +0,0 @@
...
-  members:
-    [
-      joe_chen,
-      camden_cheek,
-      keegan_carruthers-smith,
-      mohammad_umer_alam,
-      bolaji_olajide,
-      warren_gifford,
-      jacob_pleiness,
-    ]
```
craigfurman, line 14: I noticed that there is this release_guild reference under data/guilds.yml (in fact it's the only element in that list). Following a chain of greps, that file appears to be used to replace the generator directive `guild_roster`: https://github.com/sourcegraph/handbook/blob/43ba32f68284687fd0c7cb11a05ac86c53df7ac4/src/lib/markdownToHtml.ts#L354

But, I can't find `guild_roster` anywhere else in the handbook. It looks like guilds are an unused feature of the handbook. I would ask the committer of the generator code but they appear to have left the company. If the release guild has indeed been replaced by the release team, we could remove this element? WDYT?

I didn't remove the guilds feature altogether, in case another guild wants to use it!

jdpleiness: Yeah we can remove this. I'm also unsure around what `guild_roster` is for as well 🤔

jdpleiness (approved):

data/guilds.yml
DaedalusG: This was likely used in the old release tool for some generated message or something. Missed in our deprecation efforts so far. Nice catch

PR #8911, opened 2024-04-18
msp-ops: add alert information + link to alert dashboard

jac:
Adds all service alerts to the bottom of the ops pages.

Adds a link to the msp dashboards to the per environment info table

generated from: https://github.com/sourcegraph/sourcegraph/pull/61939


bobheadxi (approved):
yeet, let's go 😁

PR #8919, opened 2024-04-26
msp-ops: architecture diagrams

jac:
this handbook is still the source of truth for msp-ops pages
---

This PR adds architecture diagrams generated per environment for each service https://github.com/sourcegraph/sourcegraph/pull/62213


unknwon (approved):
WOAOOW! 😍

PR #8944, opened 2024-05-17
Update licensing.md to remove Cloud trial link

iskyOS:
Removing old Sourcegraph Cloud trial link


iskyOS:
@madisongclark This may not be mission critical -- just trying to delete a link to an old subdomain, signup.sourcegraph.com, that we're taking down. I've also removed it on the notion side.

madisongclark (approved):
Thanks for also updating in Notion, @iskyOS.

